PR #4, opened 2017-01-20
Add bastion support

metadave:
@jgmize bastion support works nicely. I chose `calico` for the network mode, be we can definitely choose something else.


metadave:
holding off until we solve our logging issues

PR #6, opened 2017-01-25
add fluentd support during install

metadave:
add git-crypt to dev node

### DIFF
--- a/k8s/install/etc/config.sh.template
+++ b/k8s/install/etc/config.sh.template
@@ -12,8 +12,8 @@ export KOPS_MASTER_ZONES="us-east-1b,us-east-1c,us-east-1d"
 export TF_STATE_BUCKET="${KOPS_SHORT_NAME}-tf-state"
 
 # populate these if installing FluentD->PaperTrail DaemonSet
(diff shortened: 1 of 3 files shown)


metadave:
note: Dockerfile + config template for fluentd located [here](https://github.com/mozmar/fluentd_k8s)

k8s/install/etc/config.sh.template
```diff
@@ -11,13 +11,18 @@ export KOPS_ZONES="us-east-1b,us-east-1c,us-east-1d"
 export KOPS_MASTER_ZONES="us-east-1b,us-east-1c,us-east-1d"
 export TF_STATE_BUCKET="${KOPS_SHORT_NAME}-tf-state"
 
+# populate these if installing FluentD->PaperTrail DaemonSet
+export PAPERTRAIL_HOST=""
+export PAPERTRAIL_PORT=""
```
jgmize: NIT: these could work with any remote syslog, perhaps we should rename these to `SYSLOG_HOST` and `SYSLOG_PORT`?

metadave: agreed, updates incoming

PR #7, opened 2017-01-26
add ssh security group tool for kops clusters

metadave:
- requires `aws` and `jq`
- optionally requires `dig` if using `--myip`
- these deps are checked at runtime and the script will print an error message if they aren't installed.

ping @jgmize

### DIFF
--- a/k8s/tools/k8s_ssh_security.sh
+++ b/k8s/tools/k8s_ssh_security.sh
@@ -1,0 +1,157 @@
+#!/bin/bash
+
+
+if [ -z "${KOPS_NAME}" ]; then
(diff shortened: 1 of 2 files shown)


k8s/tools/k8s_ssh_security.sh
```diff
@@ -0,0 +1,157 @@
...
+}
+
+for arg in "$@"
+do
+    case $arg in
+        --show)
+        SHOW=1
+        shift
+        ;;
+        --myip)
```
jgmize: This is great, but as we discussed yesterday I'd like us to also accept an IP directly. Not a blocker, a TODO

PR #11, opened 2017-01-30
use helm to install fluentd

metadave:


metadave:
see also: https://github.com/mozmar/ee-infra-private/pull/51

metadave:
updated

PR #13, opened 2017-01-31
WIP: add annotations for elb https

metadave:
- also: specify region when setting up Deis DNS


metadave:
changed moved to a different branch

PR #18, opened 2017-02-03
Deis DNS via k8s annotation, run stage2 from dir with config.sh instead of out/terraform, misc

metadave:
- instead of running stage2 from `./out/terraform`, run it from the same dir as stage1
- add a k8s annotation that creates a deis.foo.moz.works CNAME pointing to the ELB
- added a function called `install_deis` that runs all the workflow install steps
- rename `install_workflow` to `install_workflow_chart`
  - instead of parsing Terraform output with `jq`, just supply the param name to tf and it does the same thing
Ex:
`terraform output foo` gives me the value of foo, instead of having to filter json through `jq`.
- convert tabs to spaces
  - my bad, next text editor default is lame.

Tip: append `?w=1` to the URL to see the PR without whitespace changes.
  ex: https://github.com/mozmar/infra/pull/18/files?w=1

### DIFF
--- a/k8s/install/README.md
+++ b/k8s/install/README.md
@@ -52,13 +52,14 @@ kubectl get nodes
 
 ### Installing monitoring services
 
-This step installs Mig, Datadog, New Relic DaemonSets, the k8s dashboard, and Deis Workflow. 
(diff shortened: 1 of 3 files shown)


metadave:
These should be squashed upon merge!

PR #55, opened 2017-02-16
add MDN provisioning scripts

metadave:
This used to be https://github.com/mozmar/ee-infra-private/pull/63, however we'd like to keep as much of this in the public as possible.

Additions:
- README
- a `provision.sh` script that sets up an encrypted S3 bucket for Terraform remote state

fyi @jwhitlock

### DIFF
--- a/mdn/mdn_infra.tf
+++ b/mdn/mdn_infra.tf
@@ -2,6 +2,8 @@ provider "aws" {
   region = "${var.region}"
 }
 
+
+
 resource "aws_elasticache_cluster" "mdn-redis" {
     cluster_id = "mdn-redis"
     engine = "redis"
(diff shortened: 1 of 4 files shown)


mdn/mdn_infra.tf
```diff
@@ -0,0 +1,54 @@
...
+  access_policies = <<CONFIG
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Action": "es:*",
+            "Principal": "*",
+            "Effect": "Allow",
+            "Condition": {
+                "IpAddress": {"aws:SourceIp": "$${var.es-source-ip}"}
```
jgmize: As discussed, let's look into using security groups as an alternative to IPs

metadave:
* updated

metadave:
closing this in favor of a new PR that includes buckets

PR #75, opened 2017-03-01
add MDN provisioning scripts

metadave:
This PR builds on https://github.com/mozmar/infra/pull/59, however, I commented out the Redis and Elasticsearch resources until needed.

**These resources have NOT been allocated via Terraform yet.**

The following buckets are created:

- `mdn-db-storage-anonymized`
- `mdn-db-storage-production`

`mdn-downloads` already exists, but with this PR, it's now managed via Terraform.

### Notes:
- user accounts and IAM policies created in https://github.com/mozmar/ee-infra-private/pull/69
- ~we may need to rename `mdn-db-storage-anonymized` and `mdn-db-storage-production`, as AWS was timing out during bucket creation.~
- ~We may need additional permissions on the IAM policy for `mdn-downloads` to allow for logging. If this is a problem, it won't show up until I run `provision.sh`.~
- ~before applying this change, I'd like to create a backup bucket with the contents of `mdn-downloads`.~ 
   - Backup at `mdn-downloads-backup`, will be deleted after this PR is merged.

cc @jgmize @jwhitlock

### DIFF
--- a/mdn/README.md
+++ b/mdn/README.md
@@ -1,0 +1,29 @@
+# MDN infra
+
+### Requirements:
+
+- Terraform
+- Access to the IAM role `admin` role via the AWS metadata API
+- awscli
+
+
+## Usage
+
+```
+./provision.sh
+```
+
+This will setup the S3 Terraform state store if needed, and then run `terraform plan` followed by `terraform apply`.
+
+---
+
+NOTE:
+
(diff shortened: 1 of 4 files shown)


metadave:
@jgmize @jwhitlock these changes have been successfully applied

note: the `mdn-downloads` policy has one changed line, as the sample db filename changed. See the second commit in this PR for more info.

PR #89, opened 2017-03-07
Allow single AZ k8s installations

metadave:
- Deis RDS still requires 2 AZ's
- there are some manual steps included to set worker node disk sizes and enable cron tasks.
- See: https://github.com/mozmar/ee-infra-private/pull/73

### DIFF
--- a/k8s/install/README.md
+++ b/k8s/install/README.md
@@ -5,10 +5,14 @@
 - kops
 - terraform
 - awscli
+- request an AWS ACM certificate for:
+    - ${KOPS_NAME}
+    - *.${KOPS_NAME}
+    - viewsourceconf (and any others you may want on the same cert)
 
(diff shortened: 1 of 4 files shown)


k8s/install/README.md
```diff
@@ -32,6 +36,53 @@ Two different sets of results of `stage1.sh` are stored in:
...
+Run the following command:
+
+```
+kops edit ig nodes
+```
+
+A yaml file will appear in your $EDITOR, you'll need to add the following block to the `spec` section:
+
+```
+  rootVolumeSize: 500
```
jgmize: please change this to 250

metadave: good catch, fixed

jgmize:
Thanks again @metadave

PR #98, opened 2017-03-09
added cloudwatch alarm list script

metadave:
cc @jgmize @glogiotatidis


glogiotatidis:
Works nicely, thanks @metadave 🍻

PR #111, opened 2017-03-22
add AWS autoscaler installer

metadave:
- default cluster autoscaler max is 4 * KOPS_NODE_COUNT
- increases default Deis router HPA to 4 * KOPS_NODE_COUNT
- needs to be squashed


jgmize:
As we discussed earlier, we also need to adjust max_size in kops' kubernetes.tf output.

metadave:
I pushed a commit to set the `max_size` to 4 * KOPS_NODE_COUNT

k8s/install/stage2_functions.sh
```diff
@@ -260,14 +260,39 @@ config_deis_router_hpa() {
...
+        jq ".awsRegion=\"${KOPS_REGION}\"" | j2y > "${OUTPUT_FILE}"
+
+    kubectl create namespace "${AUTOSCALER_NAMESPACE}"
+    helm install stable/aws-cluster-autoscaler \
+        --name aws-cluster-autoscaler \
+        --namespace "${AUTOSCALER_NAMESPACE}" \
+        -f "${OUTPUT_FILE}"
+
+    # set ASG max size so to allow the cluster autoscaler to scale up
+    sed -i "s/max_size = $KOPS_NODE_COUNT/max_size = $DEFAULT_MAX/" ./out/terraform/kubernetes.tf
```
jgmize: This implies an additional `terraform apply` step, should we document that?

metadave: no, it should be in the stage1 and run with all the other terraform. fixed.

jgmize: even better, thanks :)

PR #127, opened 2017-03-29
Snippets Tokyo infra provisioning, multi-region support included.

metadave:
Region and other vars are passed in via environment variables in `./tokyo/provision.sh`. One minor downside is that we have to use underscores in TF variable names, as `-`'s aren't supported in env var names. **Each region maintains it's own state in a shared bucket, under a different path**. See `snippets/common.sh` for details. Snippets TF shared state goes into `s3://snippets-shared-tf-state` in `us-west-2`. 

This is all based on @glogiotatidis's work with modules 🍰

### DIFF
--- a/snippets/README.md
+++ b/snippets/README.md
@@ -2,14 +2,14 @@
 
 ### How to apply config to a single cluster
 
-```
+```shell
 cd ./tokyo
 ./provision.sh
 ```
 
 ### Adding a new cluster
 
-```
+```shell
 # from this directory
 mkdir my_snippets_region
 cp ./tokyo/provision.sh ./my_snippets_region/provision.sh
--- a/snippets/README.md
+++ b/snippets/README.md
@@ -1,0 +1,22 @@
+## Snippets
+
+### How to apply config to a single cluster
+
+```
+cd ./tokyo
+./provision.sh
+```
+
+### Adding a new cluster
+
+```
+# from this directory
+mkdir my_snippets_region
+cp ./tokyo/provision.sh ./my_snippets_region/provision.sh
+# edit ./my_snippets_region/provision.sh
+chmod 755 ./my_snippets_region/provision.sh 
+git add ./my_snippets_region/provision.sh 
+git commit
+cd ./my_snippets_region
+./provision.sh
+```
--- a/snippets/tf/cache/cache.tf
+++ b/snippets/tf/cache/cache.tf
@@ -1,0 +1,20 @@
+variable "region" {}
+
+variable "region_short" {}
+
+variable "cache_node_size" {}
+
+variable "cache_port" {}
+
+variable "cache_num_nodes" {}
+
+variable "cache_param_group" {}
+
+resource "aws_elasticache_cluster" "shared-redis" {
+  cluster_id           = "redis-shared-${var.region_short}"
+  engine               = "redis"
+  node_type            = "${var.cache_node_size}"
+  port                 = "${var.cache_port}"
+  num_cache_nodes      = "${var.cache_num_nodes}"
+  parameter_group_name = "${var.cache_param_group}"
+}
--- a/snippets/tf/cache/cache.tf
+++ b/snippets/tf/cache/cache.tf
@@ -18,7 +18,8 @@ variable "cache_security_group" {}
 
 resource "aws_elasticache_subnet_group" "shared-redis-subnet-group" {
   name       = "shared-redis-subnet-group"
-  subnet_ids = ["${split(",",var.cache_subnet_ids)}"]
+  # https://github.com/hashicorp/terraform/issues/57#issuecomment-100372002
+  subnet_ids = ["${split(",", var.cache_subnet_ids)}"]
 }
 
 resource "aws_elasticache_replication_group" "shared-redis-rg" {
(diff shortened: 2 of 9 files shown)


snippets/tf/common.sh
```diff
@@ -0,0 +1,47 @@
...
+if aws s3 ls s3://${SNIPPETS_TF_STATE_BUCKET} > /dev/null 2>&1; then
+    echo "State store already exists"
+else
+    echo "Setting up state store"
+    setup_tf_s3_state_store
+fi
+
+terraform get
+
+terraform plan
```
glogiotatidis: Maybe it's useful to plan to an outfile to ensure what's applied in the next command is actually what was planned. See https://github.com/mozmar/ee-infra-private/pull/75/files#diff-eef8e7003a9006c86f00e9bb14fc4936R59

snippets/tf/snippets.tf
```diff
@@ -0,0 +1,44 @@
...
+}
+
+module "prod-alerts" {
+  source = "../tf/alerting"
+  region = "${var.region}"
+  fqdn   = "${var.fqdn_prod}"
+  name   = "${var.alarm_name_prod}"
+}
+
+# do we want stage alerts?
```
glogiotatidis: Nope, we don't have now and I don't see us acting on those

snippets/tokyo/provision.sh
```diff
@@ -0,0 +1,18 @@
...
+export TF_VAR_region_short="tokyo"
+export TF_VAR_fqdn_prod="snippets-prod.tokyo.moz.works"
+export TF_VAR_fqdn_stage="snippets-stage.tokyo.moz.works"
+export TF_VAR_alarm_name_prod="Snippets Prod Tokyo"
+export TF_VAR_alarm_name_stage="Snippets Stage Tokyo"
+##### Redis config
+export TF_VAR_cache_node_size="cache.m4.xlarge"
+export TF_VAR_cache_port=6379
+export TF_VAR_cache_num_nodes=1
+export TF_VAR_cache_param_group="default.redis3.2"
```
glogiotatidis: Ftr we're using redis 2.8 on eu-west and us-west and I chose the same for us-east. I believe 3.2 will just work though.

metadave: When I looked in the console in `ap-northeast-1`, only `default.redis.3.2` and the clustered version were available. I originally had the value set to `default.redis.2.8`. /me shakes fist at AWS

snippets/tf/common.sh
metadave: 👍, good suggestions, I'll do that now.

metadave: fixed

snippets/tf/snippets.tf
metadave: @jgmize and I discussed this yesterday, we can have the alarms, but they won't raise any emails/IRC notifications.

snippets/tokyo/provision.sh
metadave: I think the problem here is that we need to set the redis version as well as the parameter group. The AWS console gui picks a default param group that matches the version of Redis. I'll make some updates to this PR momentarily.

metadave: ok, redis should be all set, I use an `aws_elasticache_replication_group` instead of an `aws_elasticache_cluster` and I can get 3 nodes w/ correct redis version.

metadave:
ok, redis is now created as a replication group so we get 3 nodes with correct versions. Additionally, in order to pass in a k8s node security group, I needed to pass in vpc subnets as well. Ready for another review @jgmize @glogiotatidis

snippets/tf/cache/cache.tf
```diff
@@ -0,0 +1,34 @@
...
+
+variable "cache_engine_version" {}
+
+variable "cache_subnet_ids" {}
+
+variable "cache_security_group" {}
+
+resource "aws_elasticache_subnet_group" "shared-redis-subnet-group" {
+  name       = "shared-redis-subnet-group"
+  subnet_ids = ["${split(",",var.cache_subnet_ids)}"]
```
jgmize: I know this line cost you some time figuring this out, maybe a comment with a link to the tf issue that makes this workaround necessary would be helpful for future reference?

snippets/README.md
```diff
@@ -0,0 +1,22 @@
...
+### How to apply config to a single cluster
+
+```
+cd ./tokyo
+./provision.sh
+```
+
+### Adding a new cluster
+
+```
```
glogiotatidis: mini nit: slightly better formatting if you add `shell` after the `

snippets/tf/common.sh
```diff
@@ -0,0 +1,77 @@
...
+        -backend-config="encrypt=1" \
+        -backend-config="region=${STATE_BUCKET_REGION}"
+
+    echo "Encryption for TF state:"
+    aws s3api head-object --bucket=$SNIPPETS_TF_STATE_BUCKET --key="snippets-${SNIPPETS_REGION}/terraform.tfstate" | jq -r .ServerSideEncryption
+}
+
+check_state_store() {
+    echo "Checking state store"
+    if aws s3 ls s3://${SNIPPETS_TF_STATE_BUCKET} > /dev/null 2>&1; then
```
glogiotatidis: The command will return `255` exit code if the bucket does not exist and will cause the script to exit due to `set -e`. Maybe set `set +e` before executing this?

I wonder if there's a special bash instructions to make some commands not to fail.

metadave: fixed!

PR #136, opened 2017-04-02
Update helm version (v2.2.3) and download URL patterns

jgmize:


k8s/install/stage2_functions.sh
```diff
@@ -34,8 +34,8 @@ install_helm() {
     if [ $? -ne 0 ]; then
         echo "Installing helm client"
         (cd /tmp && \
-            wget https://kubernetes-helm.storage.googleapis.com/helm-${HELM_VERSION}-linux-amd64.tar.gz && \
-            tar xzf /tmp/helm-${HELM_VERSION}-linux-amd64.tar.gz && \
+            wget https://storage.googleapis.com/kubernetes-helm/helm-v${HELM_VERSION}-linux-amd64.tar.gz && \
+            tar xzf /tmp/helm-v${HELM_VERSION}-linux-amd64.tar.gz && \
```
metadave: `$HELM_VERSION` has a `v` prefix, so `helm-v${HELM_VERSION}` may not be correct.

jgmize: good catch, fixing now

PR #161, opened 2017-04-07
ELB provisioning

metadave:
Ok, here's a rough draft. This was a bit trickier than expected, as we create multiple similar ELB's (via modules) potentially across multiple regions. The included scripts generate a .tfvars for each ELB in a particular region from a given k8s cluster. When terraform plan/apply is called, I take advantage of the `-var-file` param, which can be specified more than once for each ELB config. In the `all_elbs.tf` file, we must declare each ELB we want to create, so... we get lots of duplicated variable names. I can't think of a workaround for this at the moment, but it doesn't seem that bad as-is. One thing to note is that the generated .tfvar files are stored in git, if we rerun the config and the backend instances change, they'll be reflected here (which I think 👍 ).

Careers and snippets ELB's are created, check the AWS console for `us-east-1` to confirm configuration.  I could probably automate getting the list of subnets for a VPC in a given region in the future, but it's easy enough to work with for now. Also, we could add a variable per ELB that specifies if we want it to be created in "this" region. For example, maybe we don't want an ELB in Tokyo.

I'll writeup some docs for this on Monday.

See https://github.com/mozmar/infra/issues/159 for more info


elbs/virginia/snippets-virginia.tfvars
```diff
@@ -0,0 +1,6 @@
+snippets_elb_name = "snippets"
+snippets_subnets = "subnet-43125f6e,subnet-a699aaef,subnet-b6ceb6ed"
+snippets_instances = "i-007cad57f0c5b3157,i-01fd2b1570e510dfe,i-04c56e5c677568159,i-0981f33bdf95c3c1f,i-0f71cd35d3bb09512"
```
jgmize: we should bind to the ASG instead of directly to instances

jgmize: options:
- get ASG as output from kops tf
  - may only work for new clusters
- import existing ASGs
- create ELB without attaching instances, then use AWS CLI to bind to existing ASG

metadave: ah, yes, I missed that and it's critical.

We might not need kops/tf import at all, as the ASG name is just `nodes.virginia.moz.works`. Then, just a call to  `aws autoscaling attach-load-balancers ...`. I'd prefer not to tough the ASG's with TF if possible.

jgmize: sounds, good, option 3 it is then. something along the lines of

```sh
aws autoscaling attach-load-balancers \
      --auto-scaling-group-name $ASG \
      --load-balancer-names $ELB
```

should be fine

PR #169, opened 2017-04-13
add ACM cert request docs

metadave:
resolves https://github.com/mozmar/infra/issues/167

### DIFF
--- a/docs/requesting_a_cert.md
+++ b/docs/requesting_a_cert.md
@@ -1,0 +1,27 @@
+# Requesting an ACM certificate
+
+### (non-SRE's) If you don't have access to AWS ACM
+
+1. Create a new [Github issue](https://github.com/mozmar/infra) (or use an existing issue) and mention [@jgmize](https://github.com/jgmize) and [@metadave](https://github.com/metadave) saying that you would like to request a certificate. Please list ALL domains that you are requesting (including wildcard if needed).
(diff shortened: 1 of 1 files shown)


docs/requesting_a_cert.md
```diff
@@ -0,0 +1,27 @@
...
+
+### (SRE's) If you have access to AWS ACM
+
+1. Create a [Github issue](https://github.com/mozmar/infra) (or use an existing issue) and mention [@jgmize](https://github.com/jgmize) and [@metadave](https://github.com/metadave) saying that you would like to request a certificate. Please list ALL domains that you are requesting (including wildcard if needed).
+2. from the AWS console, select the `Certificate Manager` app
+3. click the `Request a certificate` button at the top left
+4. Enter the domain names specified in the issue from step 1.
+5. Click the `Review and request` button.
+6. Click the `Confirm and request` button.
+7. An email is now sent from AWS to the domain administrator (@jgmize). The domain admin needs to click the link in the email approving the new certificate(s).
```
jgmize: The email isn't sent just to me, it's sent to the hostmaster list that I'm on. If I'm unavailable there are folks in webops that also get the emails and can click the link if necessary. There's also an additional step that I perform after clicking the link, which is to reply to the hostmaster group that I approved it.

PR #190, opened 2017-04-21
multi-region MDN provisioning

metadave:
- the `mdn/shared` Terraform is shared across all regions (just S3 for now, maybe more in the future).
- created `mdn/multi-region` to provision EFS and any other resources we'll need.


metadave:
`shared` TF applied, MDN buckets are good to go.

metadave:
This includes 3 additional S3 buckets for MDN: `mdn-assets`, `mdn-legacy` and `mdn-legacy-samples`. 


See also:
- https://github.com/mozmar/infra/issues/197
- https://docs.google.com/document/d/1-s343yxBMiugPQm5w2ho7EcSAcaLU4uOF6p2MLgbS1I/edit#

PR #247, opened 2017-05-11
add bedrock stage and prod ELB's

metadave:
- created in Virginia and Tokyo
- ~health checks are failing for ELB's, so all instances show as `OutOfService`~. 
~The new HTTP health check is pointing at `/` using the [corresponding port](https://github.com/mozmar/infra/blob/master/elbs/tf/elbs/elbs.tf#L27). Defaults for health checks are overridable per [these tf vars](https://github.com/mozmar/infra/blob/master/elbs/tf/elbs/variables.tf#L41-L64).~
- Bedrock health check endpoints are `/healthz/`, _including a trailing slash_.


metadave:
health checks pointed at `/health-check`, similar to the existing mozilla.org ELBs.

metadave:
ELBs provisioned and working in Virginia and Tokyo.

PR #311, opened 2017-06-16
MozMEAO app automation + Frankfurt cluster, round 1

metadave:
This PR establishes a file and directory structure for each application that can be used to assist in the setup of new AWS/Kubernetes regions. I scraped our infra repo looking for commands that have been run to create apps, set config values, set scaling/limits, and added them to one of 3 scripts for each app (`setup.sh`, `scale.sh`, `teardown.sh`). The idea  is that all of our apps should have any scripting and infra install docs here. While it's not currently possible to automate all of our app installs, this PR gets us a little bit closer. For example, we can setup all infrastructure pieces, create a Deis app and Kubernetes nodeports, but a Deis config still needs to be pushed to the app. 

@jgmize @glogiotatidis @pmac would appreciate your review of [apps/README.md](https://github.com/mozmar/infra/blob/354edd200fac27ccdca8ec8a5cc53525be538bfe/apps/README.md), however I don't expect everyone to review the entire PR :-)

While the PR is large, it's mostly 0) setup/teardown/scale scripts, 1) refactoring directory structures, 2) removing unneeded .tfvars files (they're generated dynamically) and 3) ELB tweaks/new Frankfurt region.

### ELB changes

- all ELB's can specify which regions they should be created in via a map (`elbs/tf/variables.tf`). Previously, there were only a few ELB's that supported this functionality. Below is an example map to define where snippets ELB's should be created:

```
variable "snippets-elbs-by-region" {
  type = "map"
  default = {
    "ap-northeast-1" = true
    "eu-central-1" = true
    # we can leave out "us-east-1" here to skip virginia
  }
}
```

When skipping a region for a specific ELB, it's variables still need to be passed to Terraform even though they won't be used. This is the purpose of the `gen_dummy_elb_cfg` function.

For example:

    gen_dummy_elb_cfg "surveillance-prod" > $SURVEILLANCE_PROD_VARFILE

### Frankfurt cluster provisioning

This PR includes ELB creation for the new Frankfurt K8s cluster. The ELB definitions are incomplete in this PR, and will be iterated over in future PR's. Apologies for submitting as part of a monster PR, but It's Already Too Late (I worked on Frankfurt + app installation at the same time, and I don't trust my git skills enough to confidently peel away just the Frankfurt stuff).


elbs/tf/all_elbs.tf
```diff
@@ -93,6 +96,7 @@ module "careers" {
 }
```
glogiotatidis: Can you please add
` health_check_http_path = "/healthz/"` 

now that careers has a healthcheck endpoint?

elbs/tf/all_elbs.tf
```diff
@@ -73,6 +74,7 @@ module "snippets" {
...
```
glogiotatidis: Similar to careers can you please add
`
health_check_http_path = "/healthz/"
`
now that snippets has a healthcheck endpoint?

glogiotatidis:
> @jgmize @glogiotatidis @pmac would appreciate your review of apps/README.md, however I don't expect everyone to review the entire PR :-)

readme lgtm, nice work @metadave

metadave:
@glogiotatidis careers + snippets health check url updated, ok to apply in each cluster?

glogiotatidis:
> @glogiotatidis careers + snippets health check url updated, ok to apply in each cluster?

yes!

Issue #362, opened 2017-07-25
MDN interactive examples S3/Cloudfront hosting


metadave:
https://github.com/mdn/interactive-examples/issues/54#issuecomment-315131908

[Note](https://bugzilla.mozilla.org/enter_bug.cgi?product=Infrastructure%20%26%20Operations&component=WebOps%3A%20SSL%20and%20Domain%20Names) for @metadave

metadave:
let's discuss infra here:
cc @jwhitlock @schalkneethling @jgmize

metadave:
Info we'll need:
- domain
  - `moz.works` vs a "more official" sounding domain (mozilla.net?)
- ACM certificate
- IAM role automation (private)
- Cloudfront/S3 automation (public)
-

pmac:
Just a note:

On bedrock we've chosen to assign a CNAME to any `*.moz.works` resources we need for a couple reasons:

1. It's more obvious that the requests to this resource are 1st party and mozilla owned if it's a `*.mozilla.net` address.
2. You'll likely already need `*.mozilla.net` in a CSP if you have or will have one, so that's one less domain to have in the policy.

These are not deal-breakers, but are considerations for which domains to use for such things. I'll also add that we should **not** use `*.mozilla.org` for these to avoid cross-domain cookie security issues.

metadave:
@jwhitlock @jgmize @schalkneethling what domain should we use? (also, see @pmac's comment above)

schalkneethling:
There is a mdn.mozillademos.org that might be an option?

metadave:
There is also [this MDN infra issue](https://github.com/mozmeao/infra/issues/214) related to mdn.mozillademos.org.

jwhitlock:
We can't reuse ``mdn.mozillademos.org``, and shouldn't - it's already doing two things when it should just do one.

One option is a subdomain, like ``interactive-examples.mdn.mozillademos.org``, that is just this project.  WebOps still owns that domain, but they are very fast (less than 15 minutes) to add DNS entries.

Another options is something like ``interactive-examples.mdn.mozilla.net``.  I imagine it has the same issues, that WebOps owns ``mozilla.net``, but it would set a new example.  ``mdn.mozilla.net`` would be a decent source domain for server emails as well.

metadave:
[...]

PR #367, opened 2017-07-26
static hosting for interactive-examples.mdn.mozilla.net

metadave:
- restructured MDN app directory
- this PR is noisy, as I split out the main MDN infra scripts into `apps/mdn/mdn-aws` so we can accommodate other static hosting/related services.
- everything that needs review is in `apps/mdn/interactive-examples`.
- Terraform has NOT been applied.
- ~This will need an associated internal IAM PR~
  - https://github.com/mozmeao/ee-infra-private/pull/106

https://github.com/mozmeao/infra/issues/362


metadave:
provisioned!

PR #368, opened 2017-07-27
Mdn k8s makefile and YAML fixes (mostly around MySQL)

escattone:
This is another iteration (after a very frustrating day).
- I think it's simpler and we have more control (e.g., we don't have to use a storage class to match the PVC with the PV) to use our own YAML for MySQL rather than the MySQL Helm chart. It all fits easily into a single file. I removed the use of secrets, but we can easily add secrets later if we want.
- Use `kubectl apply` rather than `kubectl create`
- Stop using storage classes (since we don't have to)

I'm still seeing an issue with mounting the volume on the MySQL pod (showing partial output from `kubectl get events`):
```
12m        12m         1         mysql-622934028-4r3s3   Pod                                 Warning   FailedMount         {kubelet ip-172-20-63-17.ec2.internal}   Unable to mount volumes for pod "mysql-622934028-4r3s3_mdn-dev(50ab35bf-7321-11e7-846b-12d03c584732)": timeout expired waiting for volumes to attach/mount for pod "mdn-dev"/"mysql-622934028-4r3s3". list of unattached/unmounted volumes=[data default-token-9xdtp]
12m        12m         1         mysql-622934028-4r3s3   Pod                                 Warning   FailedSync          {kubelet ip-172-20-63-17.ec2.internal}   Error syncing pod, skipping: timeout expired waiting for volumes to attach/mount for pod "mdn-dev"/"mysql-622934028-4r3s3". list of unattached/unmounted volumes=[data default-token-9xdtp]
18m        18m         1         mysql-622934028         ReplicaSet                          Normal    SuccessfulCreate    {replicaset-controller }                 Created pod: mysql-622934028-4r3s3
20m        20m         1         mysql                   PersistentVolumeClaim               Warning   ClaimLost           {persistentvolume-controller }           Bound claim has lost its PersistentVolume. Data on the volume is lost!
```

I wasn't seeing this mount issue yesterday, so it may have something to do with the cluster or perhaps the way PV's interact with the underlying AWS EBS volumes when deleted, I don't know.


escattone:
Oops, I need to rebase this to reflect the latest directory structure changes that @metdave made. I'll do that right now.

metadave:
:+1:

per our discussion, I'd like to eventually split out environment-specific vars into separate files.

escattone:
@metadave 👍 Let's definitely go with your bash-style env var setup. I liked that. Would you like to make a PR to my branch to add a start of that, or can you send me your setup and I can modify it as needed and add it? The env variables within the `Makefile` are defaults (since the assignment is made with `?=`, they're only assigned values if not already defined outside the `Makefile`). I think the combination of your bash-style env setup and the `Makefile` should work well.

metadave:
@escattone I'll create a card to track bash-style env var changes, I won't be able to get to it until "Tuesday-ish"

escattone:
I used `make aws-create-mysql-vol` to create an AWS EBS volume (40GB, gp2) for testing purposes, and then I successfully ran `make k8s-mysql` in the Portland cluster! 🍾 The PV, PVC, service, and pod all successfully started-up, and I was able to `kc exec -i -t <pod> bash` and log into the MySQL server.

Thanks so much @metadave for creating the Portland cluster (using K8s 1.7.2, which is awesome) and getting me set-up with AWS credentials and permissions for, among other things, creating EBS volumes.

I'm going to go ahead and merge this work so we can make more iterations next week.

Issue #363, opened 2017-07-25
Move Nucleus to Frankfurt


metadave:
See also: https://github.com/mozmeao/infra/issues/184

metadave:
ping @jgmize

metadave:
related: https://github.com/mozilla/nucleus/pull/27

jgmize:
using scripts and docs in https://github.com/mozmeao/infra/tree/master/apps/nucleus to create nucleus-dev & nucleus-prod apps in frankfurt

jgmize:
https://github.com/mozmeao/infra/pull/375#issuecomment-318946168

PR #375, opened 2017-07-31
Move nucleus from virginia to frankfurt

jgmize:
For #363


jgmize:
This has been applied and DNS has been updated to point to the ELB created with this change.

elbs/tf/common.sh
```diff
@@ -28,7 +28,7 @@ get_elb_access_group_id() {
 
 check_k8s_context() {
   current=$(kubectl config current-context)
-  if [ "${current}" != "${KOPS_NAME}" ]; then
+  if [[ ! "${KOPS_NAME}" =~ "${current}" ]]; then
```
jgmize: FYI I shorten my context names and use [bash functions of the same name](https://github.com/jgmize/dotfiles/blob/master/.bash_functions.d/k8s#L29-L33) to switch between clusters

metadave:
👍

Issue #332, opened 2017-07-10
move surveillance to S3/CloudFront


metadave:

jgmize:
See https://github.com/mozmeao/infra/issues/335#issuecomment-317538890

jgmize:
Requested and approved cert in us-east-1 region for use in cloudfront.

arn:aws:acm:us-east-1:236517346949:certificate/05191cfb-620b-4d16-bdda-ac155f5fd665

jgmize:
https://github.com/mozmeao/infra/pull/376#issuecomment-318954399

PR #376, opened 2017-07-31
Move surveillance.mozilla.org to s3+cloudfront for #332

jgmize:


jgmize:
Terraform applied, DNS updated, and https://surveillance.mozilla.org/ is now being served from s3+cloudfront

metadave:
looks good 👍

Issue #374, opened 2017-07-28
MDN build automation


metadave:
Allow our MDN installation to support single-region/multi-environment (dev/stage/prod):
https://github.com/mozmeao/infra/pull/368#issuecomment-318690253

- this issue covers `k8s-services`/`k8s-deployment` from the image below.
- `aws-infra` will be tracked [here](https://github.com/mozmeao/infra/issues/378)

metadave:
cc @escattone 
click the image to make it larger.
![mdn_make_targets_xmind](https://user-images.githubusercontent.com/58244/28840068-c39c1904-76c3-11e7-934a-eeef92db1742.png)

metadave:
@escattone anything else we should include here, or do you think this issue can be closed?

escattone:
@metadave I'd like to keep this open a bit longer, to do these things:
- adjust settings
- ensure kuma, kumascript, and integration tests (at least the ones that make sense) run fine
- decide how we're going to handle the stage/prod secrets

Does that make sense, or should we create new cards for those tasks?

metadave:
works for me!

metadave:
per a discussion with @escattone, we're going to close this.

PR #377, opened 2017-07-31
reorganize env vars

escattone:
This is another iteration of the `Makefile` and the YAML files. The environment variables have been significantly reorganized.

The following commands (from the `Makefile`) seem to work:
```
make aws-create-mysql-vol
make aws-describe-mysql-vols
make k8s-ns
make k8s-shared
make k8s-delete-shared
make k8s-mysql
make k8s-delete-mysql
```

When I run `make k8s-web`, the service is created without error, but the deployment currently fails because it is unable to mount the `shared` PVC due to a connection timeout. I'm debugging that issue but I may need to pair on that with @metadave .


metadave:
@escattone I have to look in the morning, but I was having problems mounting EFS due to a timeout, guessing it's a firewall/networking issue.

escattone:
@metadave That makes sense to me as well. I commented-out the mount to check for problems further down the line, and it looks like the "web" pod is having problems connecting to the `memcached` instance as well (which may be a similar issue?). We can take a look at this together tomorrow.

FYI. I also manually loaded the sample database into the dev MySQL instance doing this:
```
cd ~/repos/infra/apps/mdn/mdn-aws/k8s
wget -N https://mdn-downloads.s3-us-west-2.amazonaws.com/mdn_sample_db.sql.gz
gunzip mdn_sample_db.sql.gz
kc exec -i mysql-3391802274-m1qg3 -- mysql --user=root --password=kuma --database=developer_mozilla_org < mdn_sample_db.sql
```

metadave:
@escattone I'll take a look at EFS and memcached this morning, I'm positive it's a networking issue. I created a [card](https://github.com/mozmeao/infra/issues/378) to tracking automating these resources.

metadave:
@escattone by the way, you can skip the intermediate `gunzip` step:

```
zcat /foo/somebackup.sql.gz | mysql -u 'username' -p db ...
```

Also, [here's](https://github.com/mozmeao/mdn-mysql) where the Kuma DB instance is built, and [here's](https://github.com/mozmeao/mdn-mysql/tree/master/mysql-demo-helper) an image that will load a mysql backup into a running MySQL container. [This script](https://github.com/mozmeao/mdn-mysql/blob/master/mysql-demo-helper/load-mdn-db.sh) waits until the database is alive (via the [`urlwait`](https://github.com/pmclanahan/urlwait) command) before attempting a restore.

metadave:
@ryan I'm recreating memcached in a different subnet, I'll have a URL shortly.

escattone:
@metadave and I discussed this verbally,so I'm going to merge this work-in-progress. The next PR will contain the work that we discussed and outlined in our meeting:
- reorganize according to https://github.com/mozmeao/infra/issues/374#issuecomment-319452658
- add `kumascript` deployment
- allow the service type to be configurable for the `api` (`ClusterIP`) and `web` (`LoadBalancer` or `NodePort`) services

PR #380, opened 2017-08-02
reorganize by infra, services, deployments

escattone:
Reorganized according to https://github.com/mozmeao/infra/issues/374#issuecomment-319452658.

The following, which represents spinning-up almost the entirety of MDN, seems to work (from a quick survey of the logs)! 🎉 
```
make k8s-services
make k8s-deployments
```


metadave:
👍

PR #388, opened 2017-08-04
update MDN dev w/ new AWS resources

metadave:
- also tag EBS volume with a `Stack` tag so it can be tracked in Resource Manager


metadave:
I'm going to just merge this as it's just env changes and it's holding me back from working on https://github.com/mozmeao/infra/pull/389

Issue #237, opened 2017-05-09
implement EFS backup for MDN


metadave:
[parent issue](https://github.com/mozmar/infra/issues/147)

analysis [here](https://github.com/mozmar/infra/issues/189)

It might be easy to include EFS backup to S3 in this as well, I'll evaluate and circle back.

metadave:
blocked on this until we have more than one K8s cluster with access to EFS.

metadave:
here's a simpler way to do replication: https://github.com/mozmar/infra/issues/189#issuecomment-301800647

metadave:
wip here: https://gist.github.com/metadave/ee186f2eb003a74f391522647866902d

metadave:
removing BLOCKED tag, as we're targeting a single region for release.

metadave:
wip [here](https://github.com/mozmeao/infra/pull/389)

metadave:
implemented [here](https://github.com/mozmeao/infra/pull/389)

PR #389, opened 2017-08-04
EFS backup tool

metadave:
- parent https://github.com/mozmeao/infra/issues/237


metadave:
@escattone ready for review!

PR #406, opened 2017-08-09
Adjust settings and fix several issues

escattone:
This PR adjusts the dev/stage/prod settings more towards what their final values should be. It also fixes several issues, including:
- the `web` and `api` deployments were not binding to the correct container port
- updates `dev.sh` with the new `redis` and `memcached` ElasticCache instances as well as EBS volume and EFS instance that @metadave created via his Terraform scripts
- sets the `DOMAIN` properly for the `dev` target environment (uses `kubectl` to extract the URL of the `web` service's load balancer)

@metadave This PR fixes the DB lock timeout issue that I've been chasing all afternoon. 🎉


apps/mdn/mdn-aws/k8s/Makefile
```diff
@@ -25,6 +25,7 @@ export MYSQL_IMAGE ?= quay.io/mozmar/mdn-mysql
 export MYSQL_IMAGE_TAG ?= latest
 export MYSQL_IMAGE_PULL_POLICY ?= IfNotPresent
 MYSQL_STORAGE_SIZE_IN_GB ?= 40
+export MYSQL_STORAGE_NAME ?= ${MYSQL_NAME}-${TARGET_ENVIRONMENT}
```
metadave: nice!

apps/mdn/mdn-aws/k8s/kuma.deploy.yaml.j2
```diff
@@ -5,9 +5,10 @@
             - gunicorn
           args:
             - "-w={{ KUMA_GUNICORN_WORKERS }}"
-            - "--bind=0.0.0.0"
+            - "--bind=0.0.0.0:{{ KUMA_CONTAINER_PORT }}"
             - "--access-logfile=-"
             - "--timeout={{ KUMA_GUNICORN_TIMEOUT }}"
+            - "--worker-class=meinheld.gmeinheld.MeinheldWorker"
```
metadave: I thought we weren't going to use Meinheld, per the SRE meeting today.

apps/mdn/mdn-aws/k8s/Makefile
```diff
@@ -118,16 +119,22 @@ export KUMA_MOUNT_PATH ?= /www
 export KUMA_DEBUG ?= False
 export KUMA_DEBUG_TOOLBAR ?= False
 export KUMA_PROTOCOL ?= "https://"
+ifeq (${TARGET_ENVIRONMENT},dev)
+export KUMA_DOMAIN=$(shell kubectl get --ignore-not-found -n ${K8S_NAMESPACE} svc ${WEB_SERVICE_NAME} -o jsonpath={.status.loadBalancer.ingress[0].hostname})
+else
```
metadave: will we eventually need a `developer.allizom.org` `KUMA_DOMAIN` for staging?

metadave:
actually, lets bump the dev requirements way day, maybe 100m for CPU, and <= 500MB for memory. If there are performance issues, we can easily increase these values.

apps/mdn/mdn-aws/k8s/kuma.deploy.yaml.j2
escattone: Yes, sorry, I'll take it out.

apps/mdn/mdn-aws/k8s/Makefile
escattone: The `regions/portland/stage.sh` setup has that defined. I'm still operating under the assumption that the env variable values in the `Makefile` are the defaults if the env variable is not defined (assuming the `dev` target environment), and that we'll use the `regions/portland/{dev|stage|prod}.sh` scripts to the the setup before using `make`.

escattone:
@metadave I'll bump all of the CPU/Memory resource requests for `dev` down to `100m`/`256Mi`.

PR #418, opened 2017-08-14
add mdn admin node for easy EFS access

metadave:
This make target creates a debian container that runs forever (until you delete it). This is great for interacting with EFS (and maybe the awscli if needed).

To connect to the running pod, first find the full name of the `mdn-admin` pod and then exec into it like this:

```
kubectl -n mdn-dev exec -it mdn-admin-3097159426-mks6l bash
```

### DIFF
--- a/apps/mdn/mdn-aws/k8s/Makefile
+++ b/apps/mdn/mdn-aws/k8s/Makefile
@@ -158,6 +158,7 @@ export BACKUP_SECRET_KEY ?= bar
 export BACKUP_ACCESS_KEY_BASE64 ?= $(shell echo -n "${BACKUP_ACCESS_KEY}" | base64)
 export BACKUP_SECRET_KEY_BASE64 ?= $(shell echo -n "${BACKUP_SECRET_KEY}" | base64)
 
+export ADMIN_NODE_NAME ?= mdn-admin
 
 aws-create-mysql-vol:
 	aws ec2 create-volume \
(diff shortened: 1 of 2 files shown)


metadave:
@escattone using this container, I'm able to connect to the container and chown an EFS directory 

```
root@mdn-admin-debian-3125158672-4976f:/mdn/www# mkdir foo
root@mdn-admin-debian-3125158672-4976f:/mdn/www# ls -la
total 32
drwxr-xr-x  7 1000 root 6144 Aug 15 13:52 .
drwxr-xr-x  3 1000 root 6144 Aug 11 13:37 ..
drwxr-xr-x  8 1000 root 6144 Aug 11 20:31 attachments
drwxr-xr-x  3 root root 6144 Aug 11 14:02 diagrams
drwxr-xr-x  2 root root 6144 Aug 15 13:52 foo
-rw-r--r--  1 1000 root   29 Aug 11 14:01 lastsync.txt
drwxr-xr-x 13 root root 6144 Aug 11 14:03 presentations
drwxr-xr-x 25 root root 6144 Aug 11 14:03 samples
root@mdn-admin-debian-3125158672-4976f:/mdn/www# chown -R 1000:1000 ./foo
root@mdn-admin-debian-3125158672-4976f:/mdn/www# ls -latr
total 32
drwxr-xr-x  3 1000 root 6144 Aug 11 13:37 ..
-rw-r--r--  1 1000 root   29 Aug 11 14:01 lastsync.txt
drwxr-xr-x  3 root root 6144 Aug 11 14:02 diagrams
drwxr-xr-x 13 root root 6144 Aug 11 14:03 presentations
drwxr-xr-x 25 root root 6144 Aug 11 14:03 samples
drwxr-xr-x  8 1000 root 6144 Aug 11 20:31 attachments
drwxr-xr-x  2 1000 1000 6144 Aug 15 13:52 foo
drwxr-xr-x  7 1000 root 6144 Aug 15 13:52 .
```

jgmize:
This works of course, but I think it might be better to do one-off commands like chowns as [jobs](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/) instead. @escattone and I discussed using initContainers for this purpose as well, but I don't think that we should be running recursive chowns as a blocker for every pod creation.

metadave:
I'm :+1: for using a job to setup ownership of EFS dirs during installation, [here's an issue](https://github.com/mozmeao/infra/issues/422) to track.

This PR still stands, it's super useful to have a utility container to be able to interact with EFS manually.

escattone:
By the way, I also experimented with the following in the kuma (`web` and `api`) deploys:
```
      initContainers:
      - name: adjust-volume-mount-ownership
        image: busybox
        command: ["sh", "-c", "chown -R 1000 {{ KUMA_MEDIA_ROOT }}"]
        volumeMounts:
        - name: kuma-fs
          mountPath: {{ KUMA_MOUNT_PATH }}
```
but decided against it (for now) for two reasons:
1. It was really slow. It added at least 2 minutes to the start-up of the container.
2. It only needed to be done once.

PR #434, opened 2017-08-17
add RDS provisioning for MDN

metadave:
This PR adds Terraform resources to create MDN stage and prod RDS instances in us-west-2. The resources are currently commented out, but I'm scheduled to create all our staging resources some time next week.

https://github.com/mozmeao/infra/issues/421


escattone:
Thanks @metadave! Perhaps I'm mistaken as to the purpose of the `mysql_db_name` variable, but I'm guessing we'll have to change that to what we use currently for both stage and prod, which is `developer_mozilla_org`.

jwhitlock:
We can change the database name when we load it from the production database dump. I'm not 100% sure we get to pick the database name in RDS, but I may be thinking of Heroku.

escattone:
From the [AWS docs (see `DBName` parameter)](http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html), it seems that we can specify the name of the database if we want one created initially.

metadave:
I'm fine if we keep things consistent, DB names updated in https://github.com/mozmeao/infra/pull/450, but I'll have to recreate the DB's.

PR #439, opened 2017-08-18
MDN elb automation tweaks / redirector service

metadave:
- add redirector service to force http->https
- add `elb_access` security group to new ELB
- we can't use some of the [ELB annotations](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go#L72-L141) here, as they apply to _all_ listeners (we have both 80 AND 443 listeners)


escattone:
Thanks @metadave! Nice work! It's disappointing though that K8s doesn't yet accept fuller annotations (for example, additional listeners on a load balancer, or per-listener access-log annotations), so that it forces us to do this kind of post-creation configuration.

It seems that in the current K8s setup for MDN, we only have one `redirector` service that serves both stage and prod? I'm wondering if we should have one each for stage and prod, so for example, we can explore changes (nginx major version updates, etc.) to the `redirector` service in staging without affecting production?

jgmize:
> Thanks @metadave! Nice work! It's disappointing though that K8s doesn't yet accept fuller annotations (for example, additional listeners on a load balancer, or per-listener access-log annotations), so that it forces us to do this kind of post-creation configuration.

Agreed that this was nice work by @metadave, and that it would be nice if this functionality was available in k8s directly. I'd say this is a good opportunity to contribute to kubernetes if you're interested.  Please let me know if that's the case; I'd be happy to help either of you (or anyone else reading this) get started on that.

> It seems that in the current K8s setup for MDN, we only have one redirector service that serves both stage and prod? I'm wondering if we should have one each for stage and prod, so for example, we can explore changes (nginx major version updates, etc.) to the redirector service in staging without affecting production?

There is generally only one redirector service per cluster, and given the simplicity of that service I would suggest that we can easily test changes like nginx upgrades independent of any particular app.

PR #442, opened 2017-08-22
tweak bedrock setup/scaling

metadave:
@jgmize we aren't setting limits for the clock proc in Bedrock, this PR proposes some new values that have *NOT* been applied to any clusters.

### DIFF
--- a/apps/bedrock/scale.sh
+++ b/apps/bedrock/scale.sh
@@ -8,22 +8,18 @@ deis limits:set web=250m/1000m --cpu -a bedrock-prod
 deis limits:set clock=300M/600M -a bedrock-prod
 deis limits:set clock=250m/1000m --cpu -a bedrock-prod
 deis autoscale:set web --min=5 --max=20 --cpu-percent=80 -a bedrock-prod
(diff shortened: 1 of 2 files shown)


metadave:
cc @pmac

apps/bedrock/scale.sh
```diff
@@ -5,17 +5,28 @@ check_meao_env
 
 deis limits:set web=300M/600M -a bedrock-prod
 deis limits:set web=250m/1000m --cpu -a bedrock-prod
+deis limits:set clock=300M/600M -a bedrock-prod
+deis limits:set clock=250m/1000m --cpu -a bedrock-prod
 deis autoscale:set web --min=5 --max=20 --cpu-percent=80 -a bedrock-prod
+#deis autoscale:set clock --min=1 --max=1--cpu-percent=80 -a bedrock-prod
```
jgmize: Why would we ever use autoscale for a singleton pod?

metadave: we wouldn't. I wasn't sure if we wanted to be explicit and tell the autoscaler that it's min/max is always 1.

PR #450, opened 2017-08-25
provision MDN stage resources

metadave:
- create a db parameter group per environment (stage/prod) to support any custom params
  - set a value for `max_allowed_packet`, as documented [here](https://mana.mozilla.org/wiki/display/websites/developer.mozilla.org+Cluster#developer.mozilla.orgCluster-Database)
- all parameters for `resource "aws_db_instance" "mdn_rds" { ... }` were sorted, and `parameter_group_name` was added.
- provisioned MDN stage database, redis/memcached clusters, EFS
- EFS is currently empty, I'll setup the directory structure when I get back!

### DIFF
--- a/apps/mdn/mdn-aws/infra/multi_region/tf/main.tf
+++ b/apps/mdn/mdn-aws/infra/multi_region/tf/main.tf
@@ -22,13 +22,13 @@ module "efs-dev" {
     nodes_security_group = "${var.nodes_security_group}"
 }
 
-#module "efs-stage" {
-#    source = "efs"
-#    efs_name = "stage"
-#    subnets = "${var.subnets}"
-#    nodes_security_group = "${var.nodes_security_group}"
-#}
-#
+module "efs-stage" {
+    source = "efs"
+    efs_name = "stage"
+    subnets = "${var.subnets}"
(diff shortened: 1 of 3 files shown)


apps/mdn/mdn-aws/infra/multi_region/tf/main.tf
```diff
@@ -105,19 +105,20 @@ module "memcached-dev" {
...
-#    mysql_instance_class = "db.t2.medium"
-#    mysql_backup_retention_days = 0
-#    vpc_id = "${var.vpc_id}"
-#}
-#
+module "mysql-stage" {
+    source = "rds"
+    # DBName must begin with a letter and contain only alphanumeric characters
+    mysql_env     = "stage"
+    mysql_db_name = "mdnstage"
```
escattone: I think `mysql_db_name` is used to create a database of that name within the mysql instance? Do we have to change this to `developer_allizom_org` to stay consistent with a dump of our current stage db (or a dump of the prod db that has been anonymized)?

metadave: +1, I'll update soonish.

Issue #445, opened 2017-08-23
Serve MDN sitemaps, other media files


jwhitlock:
The sitemaps are initiated by ``celerybeat``, generated in [background tasks](https://github.com/mozilla/kuma/blob/3618268653584ce98fdcbcde50d67eb468f69031/kuma/wiki/tasks.py#L317-L401), and written to ``MEDIA_ROOT``.  They are low traffic, but users include search crawlers, so they are important for SEO.

metadave:
What needs to be done on the infra side of things? We already have an EFS volume mounted at `/mdn` in kuma containers. Multiple pods writing the same content may end up being problematic.

jwhitlock:
Maybe nothing, but I'm not familiar with what you and @escattone did for attachments.

In SCL3, ``MEDIA_ROOT`` is the ``/media`` folder in the kuma source tree, which uses soft links to map to NFS folders and files:

* ``media``
  * ``attachments/`` -> ``NFS/attachments/``
  * ``sitemaps/`` -> ``NFS/sitemaps/``
  * ``sitemap.xml`` -> ``NFS/sitemap.xml``
  * ``humans.txt`` -> ``NFS/humans.txt``
  * ``revision.txt`` (*Created by deployments*)
  * ``kumascript-revision.txt`` (*Created by deployment process*)
  * ``robots.txt`` / ``robots-go-away.txt`` (*Chosen by Apache config*)
  
Looking at the [settings](https://github.com/mozmeao/infra/blob/269fbd80139e0c058bbf32e3edbf3857fc1ca056/apps/mdn/mdn-aws/k8s/Makefile#L119), you are pointing ``MEDIA_ROOT`` at the attachments share. If that share has an attachments subfolder (``/mdn/www/attachments/attachments``), then there is probably nothing to do. The sitemap will be written to ``/mdn/www/attachments/sitemap.xml`` and be happy.  If that's where the attachments are stored, we'll need another process or share for those files.

metadave:
The current directory structure for EFS is as follows:

```
./mdn/
|-- test
|   |-- attachments
|   |-- diagrams
|   |-- presentations
|   `-- samples
`-- www
    |-- attachments
    |   |-- 2012
    |   |-- 2013
    |   |-- 2014
    |   |-- 2015
    |   |-- 2016
    |   `-- 2017
    |-- diagrams
    |   `-- workflow
    |-- presentations
    |   |-- eich-ajax-experience-2007
[...]

PR #455, opened 2017-08-30
update KUMA_MEDIA_ROOT default, stage, and prod

escattone:
This is one part of the work laid out in https://github.com/mozmeao/infra/issues/445#issuecomment-325834608. It simply updates the default (`Makefile`), stage, and prod settings for `KUMA_MEDIA_ROOT` to reflect the setting in dev.


jwhitlock:
The change looks good to me, but I'm not sure if I'm the one to review this.

According to @metadave's [directory listing](https://github.com/mozmeao/infra/issues/445#issuecomment-324384879), this puts ``diagrams``, ``presentations``, and ``samples`` in ``KUMA_MEDIA_ROOT``. Does that mean the files will be available at two different paths? Or does this replace other solutions for serving those files?

escattone:
@jwhitlock Good question. My understanding is that for AWS deployments they will not be served from both paths, since we'll be setting `SERVE_MEDIA` to false and all of the other files within `KUMA_MEDIA_ROOT` and `KUMA_LEGACY_ROOT` will be served with dedicated endpoints.

metadave:
apologies @escattone, I missed this PR upon returning from vacation!

PR #458, opened 2017-08-31
Enable gzip to fix #443

### LINKED ISSUE #443: gzip https://github.com/mdn/interactive-examples/issues/221

jgmize:
Also fix https://github.com/mdn/interactive-examples/issues/221


jgmize:
This change has been applied, but may take about 15 minutes to propagate per https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html

Issue #463, opened 2017-09-07
create K8s secrets resources for stage/prod


metadave:
we'll need to split out secrets so they don't need to be deployed with each Jenkins deployment.

metadave:
@escattone this just hit me this morning, we'll need to create secrets for all the stage/prod environment vars so we can create those once and then let Jenkins push new deployments.

escattone:
@metadave That's a good point! It doesn't make sense to have to deploy the secrets each time.

metadave:
I'm thinking of how Jenkins will work, I don't think we can have an unlocked mdn-k8s-private repo on the Jenkins server to populate the secrets on each push. The secrets can be stored and deployed once from mdn-k8s-private.

metadave:
https://github.com/mozmeao/infra/pull/467

PR #465, opened 2017-09-08
remove k8s-based mysql && use new private env vars in stage

escattone:
This PR removes the old k8s-based mysql service, deployment, and setup env vars, as well as uses the new private env vars for staging (see https://github.com/mozmeao/mdn-k8s-private/pull/6).


escattone:
I haven't tried these yet, but wanted to open for comment.

metadave:
How will this work for dev instances?

metadave:
regarding stage secrets: https://github.com/mozmeao/infra/issues/463

escattone:
Thanks @metadave. I'm not going to keep the "private env vars" idea since we're going to one-time deploy a k8s secrets resource and use that instead, but I'll merge this to keep the mysql-related clean-up and we'll iterate from there.

PR #467, opened 2017-09-08
use secrets (and remove mysql YAML files)

escattone:
This PR configures the deployments to use the new `mdn-secrets` resource, and also removes the MySQL PV, PVC, and deployment YAML files. We may find that we'd like to deploy a local MySQL instance for dev, but we can do that later if desired.


metadave:
nice! :+1:

PR #468, opened 2017-09-08
remove unused vestiges of the local mysql era

escattone:


escattone:
I'm going to just merge this without review, since it's simply removing unused cruft.

Issue #409, opened 2017-08-11
maintenance mode release plan


metadave:
I'm starting a list of things to take care of during the maintenance deploy:

- ensure cluster is appropriately scaled
  - ensure autoscaler max is set higher than the default (the default for the portland cluster will be very low, as it was created with 2 nodes initially).
- ensure static files from `/data` are synced over to the appropriate locations.
- disable Kuma web services
  - what's the preferred method here? via DNS?
  - do we redirect traffic to a maintenance page? 
- wait for redis/memcache queues to drain
  - what tool can we use to monitor this?
- ~backup production MySQL database~ -> we're using the anonymized db
- ~restore production database in AWS deployment~
- ensure [mdn.mozillademos.org DNS](https://github.com/mozmeao/infra/issues/214) is correct.
- change developer.mozilla.org DNS to point at AWS deployment

DNS change info: https://github.com/mozmeao/infra/issues/396

jwhitlock:
The database needs to be in place at the start of maintenance mode.  We'll need to backup the MySQL database, anonymize or encrypt it, get it to AWS, and load it.  We'll then do the DNS switch to send traffic to AWS instead of SCL3.

I think SCL3 will just not have any traffic in MM, with all the traffic going to AWS via a DNS switch.  I was thinking it would work the same way in AWS, with the "live cluster" not getting traffic, and the "mm cluster" getting the traffic, via DNS.  This allows us to do the maintenance on the live cluster, run tests using private DNS names, etc., before switching the DNS back.

Maintenance mode, in general, will have yesterday's data. Once we're in AWS, we can tighten this up with last-hour's snapshot for planned maintenance, but 24-hour-old data is better that our current "we're working on it" splash page.

It is not OK to ever have a de-anonymized production database in a publicly accessible location. We currently restrict the anonymized database to select staff and NDA-ed contributors.  I'd like to ensure we do the same with the AWS transition, by encrypting the database dump in flight and at rest, and ensuring narrow permissions on S3 buckets holding it.

metadave:
See https://github.com/mozmeao/infra/issues/401 for DB backup scripting.

jwhitlock:
I think this belongs in the A3 milestone, when we start sending production traffic to AWS in maintenance mode.

jwhitlock:
Since this was opened, we decided to do a mixed maintenance mode test:

* SCL3 will run in maintenance mode
* AWS production will run in maintenance mode on a recent production backup
[...]

PR #469, opened 2017-09-09
Start checklist for checking maintainance mode

jwhitlock:
Related to issue #409, start on a checklist for manual verification of maintenance mode in AWS.  I think a related checklist is needed for provisioning MM, which has been started in that issue, and may be useful after the prototype.

### DIFF
--- a/apps/mdn/mdn-aws/docs/mdn_maintenance_mode_checklist.md
+++ b/apps/mdn/mdn-aws/docs/mdn_maintenance_mode_checklist.md
@@ -1,0 +1,84 @@
+Here's some things to check when MDN is in maintenance mode.
+
+Note: Replace https://mdn-mm.moz.works with the base URL you are testing
+
+Base URL: https://mdn-mm.moz.works
+
+# Preparing for Maintenance Mode
+
+If you want a public-facing site, you'll need to generate
+content before putting the site in Maintenance Mode. The
+full steps are in *todo*, but here's the quick version:
+
+* [ ] Load a recent (anonymized) backup
+* [ ] Enable GitHub auth (see [Enable GitHub Auth](https://kuma.readthedocs.io/en/latest/installation.html#enable-github-auth-optional))
+* [ ] Render unrendered pages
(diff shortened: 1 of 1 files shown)


metadave:
@jwhitlock can you move these to a (new) `apps/mdn/mdn-aws/docs` directory?

jwhitlock:
I've moved the checklist to ``apps/mdn/mdn-aws/docs/mdn_maintenance_mode_checklist.md``, and went ahead with the current URL https://mdn-mm.moz.works in the docs.

I'm not sure how many manual steps are needed to get a deployment into maintenance mode - that may deserve it's own documentation.

My next step is to flesh out the manual testing section, by surveying the endpoints in the Kuma code.

apps/mdn/mdn-aws/docs/mdn_maintenance_mode_checklist.md
```diff
@@ -0,0 +1,134 @@
...
+* Interactive Examples: https://interactive-examples.mdn.moz.works
+
+# Preparing for Maintenance Mode
+
+If you want a public-facing site, you'll need to generate
+content before putting the site in Maintenance Mode. The
+full steps are in *todo*, but here's the quick version:
+
+* [ ] Load a recent (anonymized) backup
+* [ ] Enable GitHub auth (see [Enable GitHub Auth](https://kuma.readthedocs.io/en/latest/installation.html#enable-github-auth-optional))
```
safwanrahman: In maintenance mode, user cant log in. So Github auth is required?

jwhitlock: I think you need to be logged in and use the Django admin to generate the search index, so you'll need to do some things before switching to MM.

safwanrahman: In the anonymized databse, the email is anonymized also. So if any admin tries to log in with github, he will be unable to do.
Moreover, I think admin should able to login with username password in the `/admin` url!

PR #471, opened 2017-09-11
provision prod resources, enable multi-AZ RDS for MDN

metadave:
I'm waiting for a anon DB import to complete, and I'll re-provision stage once that's complete.


metadave:
I'm going to merge this as-is in order to iterate on some new changes.

PR #472, opened 2017-09-11
Added MDN Architecture Wireframes

bensternthal:


bensternthal:
Merging since this is just wireframes.

Issue #477, opened 2017-09-12
Add env overrides for SCL3 production settings (bug 1308630)


jwhitlock:
Add environment overrides for required settings in ``settings_local.py`` in SCL3, so that they can be set in AWS as well.

The SCL3 security sensitive settings are contained in a ``settings_local.py`` file that is the default Django settings file in production.  It loads settings (``from settings import *``), and then overrides and adds additional settings for production.

The files can be accessed from the ``developeradm`` admin node at:

* Production - ``/data/developer/src/developer.mozilla.org/kuma/settings_local.py``
* Staging - ``/data/developer-stage/src/developer.allizom.org/kuma/settings_local.py``

Due to the sensitive nature of these settings, I'll have to be vague about the values.

This environment override needs to be fixed:

* [x] ``ADMINS`` ([docs](https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-ADMINS)) - a list of (name, email) pairs for administration emails. It is set to ``[('MDN devs', 'mdn-dev@mozilla.com')]`` in stage and prod.  The default override sets it to a list, which I believe will break if an email is sent. For [browsercompat](https://github.com/mdn/browsercompat/blob/e283cda792293c8d91f137d4ab2ae9856b024e07/wpcsite/settings.py#L191-L193), I split these into ``ADMIN_NAMES`` and ``ADMIN_EMAILS`` and joined them in ``settings.py``.

These environment overrides need to be added, and are OK in a non-secret file:

* [x] ``CELERYD_MAX_TASKS_PER_CHILD`` ([celery docs](http://docs.celeryproject.org/en/latest/userguide/configuration.html#worker-max-tasks-per-child) and [New Relic docs](https://docs.newrelic.com/docs/agents/python-agent/back-end-services/python-agent-celery#troubleshooting)). This setting causes the ``celeryd`` child process to terminate after running a certain number of tasks. This is set to 3000 in stage and 100 in production, and the default is ``None`` for unlimited. This was a common setting to deal with memory leaks in ``celeryd`` back in 2013 or so. I'd like to try leaving at the default of unlimited, but we may need to set it based on testing.
* [x] ``DEFAULT_AVATAR`` - ``STATIC_URL + 'img/avatar.png'`` in dev, `` MEDIA_URL + 'img/avatar.png'`` in production. Maybe. This is the external URL sent to the gravatar server.
[...]

PR #480, opened 2017-09-14
add new env overrides per infra issue 477

escattone:
This PR adds some new environment variables per https://github.com/mozmeao/infra/issues/477. It depends on https://github.com/mozmeao/mdn-k8s-private/pull/14 and a Docker image for Kuma created after merging https://github.com/mozilla/kuma/pull/4426.


metadave:
@escattone should the email addresses be secrets instead (so they don't get spammed?). If they're already public, that's fine too. Feel free to merge!

jwhitlock:
@escattone and I talked about this, and neither of us have a strong opinion on email secrecy.  The ``@developer.mozilla.org`` emails are "fake" emails, and don't receive email.  This may effect delivery, and we may need to change that at some point.  The ``ADMINS`` email is real, and is already in the kuma source. 🤷‍♂️

If spam or other considerations are a concern in the future, then we will probably want the new values to be secret. So, if it is not much trouble, let's make the emails a secret setting now, but keep the known values and defaults.

escattone:
Thanks @metadave and @jwhitlock! I'll update this PR based on https://github.com/mozmeao/mdn-k8s-private/pull/17.

metadave:
🍰

PR #494, opened 2017-09-15
use single env EMAIL_URL specified from email-url secret

escattone:
The email settings were not working, and I discovered that Kuma was already coded (in `kuma/settings/common.py`) to consume the `EMAIL_URL` env override (parsed using the `dj-email-url` Python package) rather than the separate `EMAIL_HOST`, `EMAIL_HOST_USER`, and `EMAIL_HOST_PASSWORD` env overrides This PR depends on https://github.com/mozmeao/mdn-k8s-private/pull/18 which creates the `email-url` secret.


escattone:
This change has been deployed to staging and verified.

PR #496, opened 2017-09-16
initial commands to handle rolling updates for kuma & kumascript

escattone:
This PR adds the commands `make k8s-kuma-rollout` and `make k8s-kumascript-rollout` for performing rolling updates (see https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#updating-a-deployment) when the Kuma and/or Kumascript Docker images are updated.

### DIFF
--- a/apps/mdn/mdn-aws/k8s/Makefile
+++ b/apps/mdn/mdn-aws/k8s/Makefile
@@ -182,6 +182,28 @@ k8s-deployments: k8s-web k8s-api k8s-kumascript k8s-celery
 k8s-delete-deployments: k8s-delete-web k8s-delete-api \
  						k8s-delete-kumascript k8s-delete-celery
 
+k8s-kuma-rollout:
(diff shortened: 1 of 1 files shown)


escattone:
Examples of current usage:
```
export KUMA_IMAGE_TAG=dcce68f
make k8s-kuma-rollout

export KUMASCRIPT_IMAGE_TAG=d279191
make k8s-kumascript-rollout
```

metadave:
does the infra repo need to be checked out on the Jenkins server to push updates to k8s?

escattone:
@metadave Yes, I was thinking that we would check out the infra repo on the Jenkins server, but I suppose we could instead put these rollout commands into the Kuma and Kumascript `Makefiles`? Maybe that makes more sense?

metadave:
@escattone my 2 cents: let's keep it simple for now, we can leave scripts in this repo and move them if needed after the migration.

PR #499, opened 2017-09-18
use newer certs that include {prod,stage}.mdn.moz.works

metadave:


metadave:
note: the stage ELB is currently configured to use the cert for `mdn-mm.moz.works`

PR #504, opened 2017-09-19
Update snippets-prod autoscale.

glogiotatidis:


apps/snippets/scale.sh
```diff
@@ -5,6 +5,6 @@ check_meao_env
 
 deis limits:set web=300M/600M -a snippets-prod
 deis limits:set web=250m/1000m --cpu -a snippets-prod
-deis autoscale:set web --min=5 --max=15 --cpu-percent=80 -a snippets-prod
+deis autoscale:set web --min=15 --max=30 --cpu-percent=80 -a snippets-prod
 
-deis scale clock=1 -a snippets-prod
+echo "Run `deis scale clock=1 -a snippets-prod` on the cluster with db write rights"
```
glogiotatidis: Not sure if that's the best way to add this. Happy to implement different if you have better ideas

glogiotatidis: Actually `clock` should be run where `snippets-admin` runs. Should I create a new app under `apps/snippets-admin`?

This app will share resources (s3, redis, etc) with `apps/snippets` though

jgmize: it might make more sense to put it under apps/snippets/admin

metadave: @glogiotatidis do you prefer to create a new dir or keep the clock stuff in `apps/snippets`? My vote is whatever makes it easier and more apparent which commands need to be run. Thanks!

apps/snippets/scale.sh
```diff
@@ -5,6 +5,4 @@ check_meao_env
 
 deis limits:set web=300M/600M -a snippets-prod
 deis limits:set web=250m/1000m --cpu -a snippets-prod
-deis autoscale:set web --min=5 --max=15 --cpu-percent=80 -a snippets-prod
-
-deis scale clock=1 -a snippets-prod
+deis autoscale:set web --min=10 --max=30 --cpu-percent=80 -a snippets-prod
```
glogiotatidis: Changed this to `min=10` from `min=15` which reflects what's on the servers now after recent optimization changes.

apps/snippets/scale.sh
glogiotatidis: Add an `admin` directory with `setup.sh` and `scale.sh`

PR #505, opened 2017-09-19
add chown of /mdn/www dir for sitemaps

escattone:


escattone:
@metadave I don't think so, although it probably wouldn't hurt to change the ownership of the other subdirs (diagrams, samples, presentations). For now, let's try to get by without.

Issue #498, opened 2017-09-16
Manually test MM in AWS


jwhitlock:
Copied from PR #469:

This checklist uses these URLs:

* Base: https://mdn-mm.moz.works
* CDN: https://mm-cdn.mdn.moz.works
* Demo: https://mdn-mm-demos.moz.works
* Interactive Examples: https://interactive-examples.mdn.moz.works

# Preparing for Maintenance Mode

If you want a public-facing site, you'll need to generate content before putting the site in Maintenance Mode. The full steps are in *todo*, but here's the quick version:

* [x] Load a recent (anonymized) backup
* [x] Enable GitHub auth (see [Enable GitHub Auth](https://kuma.readthedocs.io/en/latest/installation.html#enable-github-auth-optional))
* [x] Render unrendered pages
* [x] Add https://mdn-mm-demos.moz.works to constance ``KUMA_WIKI_IFRAME_ALLOWED_HOSTS``
* [x] Generate sitemaps
* [x] Generate the search index
* [x] Restart web servers, etc. in Maintenance Mode

# Manual Sanity Check

## Home page

Load https://mdn-mm.moz.works/en-US

* [x] Loads without errors
* [x] Has banner "MDN is currently in read-only maintenance mode. Learn more."
* [x] No login option or view profile at the top
* [x] Has entries for the Hacks Blog

## Article page

Load https://mdn-mm.moz.works/en-US/docs/Web/HTML

* [x] Loads without errors
* [x] Has banner "MDN is currently in read-only maintenance mode. Learn more."
* [x] No login option or view profile at the top

## Maintenance Mode page

Click a banner to load https://mdn-mm.moz.works/en-US/maintenance-mode

* [x] Page is the destination of the "Learn more" link in the banner
* [x] Loads without errors

# Automated Checks

With a Kuma environment, [run the functional tests](https://kuma.readthedocs.io/en/latest/tests-ui.html), using a command like:

```
py.test --maintenance-mode tests/functional tests/redirects --base-url https://mdn-mm.moz.works --driver Chrome --driver-path /path/to/chromedriver
```

* [x] Functional tests pass

These check that disabled endpoints redirect to the maintenance page, and that other pages are OK. It duplicates the Manual Sanity Check.

# Full Manual Tests

*Note: Many of these are candidates for headless testing*

* [x] https://mdn-mm-demos.moz.works/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color - 200, sample as a stand-alone page
[...]

PR #509, opened 2017-09-20
Add an MDN initialization checklist

jwhitlock:
Add an initialization checklist with steps from setting up the maintenance mode test.

Also standardize on dashes, not underlines.

I'll add more detail later, but I wanted to capture steps from #498 as we encounter them.

### DIFF
--- a/apps/mdn/mdn-aws/docs/init-checklist.md
+++ b/apps/mdn/mdn-aws/docs/init-checklist.md
@@ -1,29 +1,128 @@
 When initializing a new MDN deployment, there are several items that may need
 to be configured.
 
-# Database
+# Set Deployment Environment
 
-* Sample DB or Anon DB or Prod Backup
+The environment settings for a deployment may need to be customized for the
+desired domain name and use cases. See the existing environments for ideas
+on what may need to be changed.
 
-# Configuration
+* [ ] Customize the deployment settings
 
-* GitHub Auth (optional, see [Enable GitHub Auth](https://kuma.readthedocs.io/en/latest/installation.html#enable-github-auth-optional))
-* Site
-* Constance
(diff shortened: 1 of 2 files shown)


apps/mdn/mdn-aws/docs/init-checklist.md
```diff
@@ -0,0 +1,128 @@
...
+  different, or you may want different values for your testing environment.
+
+# Populate Search (Optional)
+
+The database may include indexes, but these aren't reflected in the new
+deployment's ElasticSearch instance. Use the Django admin and the
+[development instructions](https://kuma.readthedocs.io/en/latest/elasticsearch.html#indexing-documents)
+to create a new index.
+
+* [ ] Create and populate index
```
escattone: Maybe change to "Create, populate, and promote index"?

apps/mdn/mdn-aws/docs/init-checklist.md
```diff
@@ -0,0 +1,128 @@
...
+
+* [ ] Load the desired database
+* [ ] (Optional) Run anonymization script
+* [ ] (Optional) Run anonymization confirmation script
+
+# Configure the Database
+
+* [ ] Set the site name:
+  ```
+  ./manage.py set_default_site --name=site-name.moz.works --domain=site-name.moz.works
```
escattone: This is a great catch, thanks! I discovered this the hard way, but now I know better.

jwhitlock: It wasn't on my mental list of stuff to setup until you mentioned it, so thanks for "discovering" it.  I wasn't aware of the ``./manage.py set_default_site`` until I looked for it. In the next month, we can make all of these management commands, and eliminate setting up an manual admin.

PR #512, opened 2017-09-20
adjust prod settings & use kuma image with robots fix

escattone:
This PR simply adjusts the prod settings, and sets the Kuma image to the one that contains https://github.com/mozilla/kuma/pull/4438.


metadave:
👍

PR #528, opened 2017-09-24
Disallow robots everywhere but developer.mozilla.org

jwhitlock:
Only the production server should have ``ALLOW_ROBOTS=True``, to protect against search engines indexing the site.

I'd like this applied as soon as possible.


escattone:
I'll deploy these imediately as well.

escattone:
Deployed to https://stage.mdn.moz.works (see https://stage.mdn.moz.works/robots.txt).

PR #531, opened 2017-09-26
remove unused MDN buckets

metadave:
This PR removes the following unused S3 buckets:

- mdn-assets
- mdn-legacy
- mdn-legacy-samples

Terraform has NOT yet been applied.


jwhitlock:
Do you think we'll be able to use these bucket names in a few months if we delete them today? Or will AWS prevent us from using an S3 name that was ever in use?

metadave:
@jwhitlock I've reused "old" bucket names in the past, and even if these are taken for some reason, we can just pick new names (`mdn-public-assets` etc).

metadave:
Since versioning has been enabled on the buckets, I'm manually deleting logs in the buckets with Python/boto3 code from [this](https://stackoverflow.com/questions/29809105/how-do-i-delete-a-versioned-bucket-in-aws-s3-using-the-cli) post:

```
import boto3
s3 = boto3.resource('s3')
bucket = s3.Bucket('your-bucket')
bucket.object_versions.all().delete()
```

metadave:
buckets deleted.

Issue #404, opened 2017-08-08
decommission Portland Deis 1 cluster


metadave:
- notify teams for each hosted app that Jenkins will need to be updated to skip pushing to usw

jgmize:
Created [VPC peering connection](http://docs.aws.amazon.com/AmazonVPC/latest/PeeringGuide/create-vpc-peering-connection.html#create-vpc-peering-connection-local) to allow nodes in new k8s VPC to talk to RDS, elasticache, etc in existing VPC.

glogiotatidis:
> Created VPC peering connection to allow nodes in new k8s VPC to talk to RDS, elasticache, etc in existing VPC.

Thanks Josh. I still can't connect to ee-mariadb and ee-redis from portland.moz.works. Is there something 
additional i should be doing?

PR #532, opened 2017-09-27
Careers Updates

glogiotatidis:
The new recourse limits have been applied to frankfurt, portland and tokyo.
New NR Synthetics monitor has been created.

 Part of #404


apps/careers/setup.sh
```diff
@@ -14,6 +14,11 @@ setup_monitors() {
         "Careers Frankfurt" \
         "https://careers-prod.frankfurt.moz.works" \
         "AWS_EU_CENTRAL_1"
+
+    create_monitor_if_missing \
```
metadave: there should be an entry in `teardown.sh` to delete the monitor as well.

glogiotatidis:
Part of #539

PR #537, opened 2017-09-28
[snippets-stats] Update configuration.

glogiotatidis:
- Reduce autoscale to min 3. Under normal load 3 per cluster are enough.
 - Removed ALLOWED_HOSTS configuration, not a Django app.
 - Add Portland NR Synthetics monitor.

Part of #538


metadave:
:+1:

Issue #543, opened 2017-09-29
Deploy cloudflare-datadog-sync to Portland


glogiotatidis:
- [x] Create app
- [x] Stop app in deis us-west
- [x] Update deploy scripts
- [x] Scale app to cmd=1
- [x] Create travis deis user
- [x] Allow travis user to deploy deis app
- [x] Test app deployment
- [x] Verify that datadog receives data
- [x] Make Deis app non-routable
- [x] Update mozmeao/infra

Part of #404

glogiotatidis:
Done!

PR #544, opened 2017-09-29
Add cloudflare-datadog-sync app configuration.

glogiotatidis:
See also #543


metadave:
thanks so much @glogiotatidis!

PR #549, opened 2017-09-29
New Relic: Labels, don't rollup KS, don't track celerycam/beat

jwhitlock:
* Update Kuma tag to current master
* New Relic doesn't merge NodeJS and Python apps. Leave KumaScript out of the rollup ``kuma-*-portland`` app, and rename the remaing app from ``kuma-kumascript-prod-portland`` to ``kumascript-prod-portland``.
* We're not tracking ``celerybeat`` (which just schedules tasks) or ``celerycam`` (which updates the database with task data) with New Relic in SCL3, so don't bother in k8s.
* Set the labels Project:MDN and Cluster:mdn-prod, to make it a little easier to filter apps in New Relic.

**Update**: Also combined ``api`` and ``celery`` into ``kuma-backend-prod-portland`` app.


metadave:
👍

PR #557, opened 2017-10-03
update prod (and stage) configuration files

escattone:
- Create a new `prod.mm.mdn.moz.works.sh` configuration (for MM traffic tests)
- Change `prod.mdn.moz.works.sh` configuration for `prod.mdn.moz.works` testing
- Change `prod.sh` configuration to reflect current understanding of "all-in" AWS deployment
    - We may want to use `cdn.mdn.mozilla.net` instead of `developer.cdn.mozilla.net`
    - Do we want to stay with `mdn.mozillademos.org`?


jwhitlock:
I like the idea of changing ``mdn.mozillademos.org``, but we'll probably need to support both this and whatever new name we use for a while. Rendered pages have the URL hard-coded in the content, so we'll have to re-render all the pages to apply the new domain name. Even then, there may be some in-content links (not generated by KumaScript) that will have to be found and fixed.

One way to transition would be an nginx instance or similar that redirects traffic from one domain to the other.

metadave:
+1 to sourcing a common set of props, but as it's not going to live for long, it's fine for now.

`mdn.mozillademos.org` is a CNAME to `mdn-demos.moz.works`, I don't see a benefit to changing it, but I don't have a strong opinion on this.

Merge away.

PR #566, opened 2017-10-05
remove outdated information from apps/mdn/mdn-aws/README.md

escattone:
The outdated information in the `apps/mdn/mdn-aws/README.md` was bothering me, so I removed it.


metadave:
+1

PR #569, opened 2017-10-09
change imagePullPolicy for MDN cronjobs

metadave:
quay.io had a minor outage this morning, which prevented the cronjobs from starting. Instead of always pulling a new image, only pull if the image is not present. We'll have to get better about tagging images in the future, but I think the `latest` tag is ok for now.

SCL3 and prod cronjobs updated in the `mdn-prod` namespace.

### DIFF
--- a/apps/mdn/mdn-aws/k8s/Makefile
+++ b/apps/mdn/mdn-aws/k8s/Makefile
@@ -144,7 +144,7 @@ export SYNC_REMOTE_DIR ?= "/backups/efs/"
 
 export BACKUP_CONTAINER_NAME ?= mdn-backup
 export SYNC_FROM_S3_CONTAINER_NAME ?= mdn-sync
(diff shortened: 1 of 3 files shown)


escattone:
@metadave I wonder if the fact that our `BACKUP_IMAGE` is [using the `latest` tag](https://github.com/mozmeao/infra/blob/54b29db8d34afd6e43475cb115d41ea65ba1a031/apps/mdn/mdn-aws/k8s/Makefile#L147) will cause the configured image-pull policy to be overridden, and always pull? The K8s docs seem to suggest that (https://kubernetes.io/docs/concepts/containers/images/#updating-images).

metadave:
good catch @escattone, fixed

metadave:
`mdn-prod` cronjobs have been updated to use the hash instead of `latest`.

PR #570, opened 2017-10-09
config elb idle timeout during service creation

metadave:
@escattone per our pairing session on Friday

### DIFF
--- a/apps/mdn/mdn-aws/k8s/configure_elb.sh
+++ b/apps/mdn/mdn-aws/k8s/configure_elb.sh
@@ -58,6 +58,14 @@ add_logging() {
         --region ${AWS_REGION}
 }
 
+configure_elb_timeout() {
+    echo "Setting ELB idle timeout"
+    aws elb modify-load-balancer-attributes \
+        --load-balancer-name ${ELB_NAME} \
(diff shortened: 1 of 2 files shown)


apps/mdn/mdn-aws/k8s/configure_elb.sh
```diff
@@ -58,6 +58,14 @@ add_logging() {
         --region ${AWS_REGION}
 }
 
+configure_elb_timeout() {
+    echo "Setting ELB idle timeout"
+    aws elb modify-load-balancer-attributes \
+        --load-balancer-name ${ELB_NAME} \
+        --load-balancer-attributes "{\"ConnectionSettings\":{\"IdleTimeout\":120}}" \
```
escattone: @metadave It would be nice (but of course not necessary) to somehow automatically keep this value in sync with `WEB_GUNICORN_TIMEOUT`. I wonder if it would be worthwhile to re-run `configure_elb_timeout()` whenever we re-deploy, just in case `WEB_GUNICORN_TIMEOUT` changes?

metadave: I've pushed an update that uses `WEB_GUNICORN_TIMEOUT` to set the ELB timeout, as well as a comment above the `WEB_GUNICORN_TIMEOUT` definition mentioning that the ELB needs to be updated if the value changes.

PR #603, opened 2017-10-13
 tf for creating MDN Frankfurt maint mode resources (without RDS) 

metadave:
- changes to backup window and maintenance window were already set via the console the day after the MDN migration, updated here to get tf back in sync.


metadave:
tf is now being applied

PR #604, opened 2017-10-13
update kuma/kumascript image tags and replicas

escattone:


metadave:
my bad, should have waited for jwhitlock to review!

PR #608, opened 2017-10-16
update prod configs

escattone:
- update API replicas from 4 to 2 (to reflect current deployment)
- removed unused production configurations


escattone:
Good catch @metadave, will do!

metadave:
:+1:

PR #609, opened 2017-10-16
update stage.sh and add stage.mm.sh

escattone:
I basically made `stage.mdn.moz.works.sh` (along with some updates) the new `stage.sh` and then added `stage.mm.sh` for putting stage into maintenance-mode.


apps/mdn/mdn-aws/k8s/regions/portland/stage.sh
```diff
@@ -28,7 +29,7 @@ export WEB_SERVICE_TYPE=LoadBalancer
 export WEB_SERVICE_PORT=443
 export WEB_SERVICE_TARGET_PORT=8000
 export WEB_SERVICE_PROTOCOL=TCP
-export WEB_SERVICE_CERT_ARN=arn:aws:acm:us-west-2:236517346949:certificate/ddf39755-b0e2-4bfa-92e5-7f2d6025db14
+export WEB_SERVICE_CERT_ARN=arn:aws:acm:us-west-2:236517346949:certificate/536f299c-ebcb-457f-9ce2-6c98a5132c82
```
jwhitlock: I didn't expect this to change. Does this change the SSL cert, or just set it to the expected value?

escattone: Good question, as it looks suspicious I know, but it simply sets it to the expected value that it was set to in the now-removed `stage.mdn.moz.works.sh` file.

metadave: The ARN ` arn:aws:acm:us-west-2:236517346949:certificate/ddf39755-b0e2-4bfa-92e5-7f2d6025db14` has the following entires:

- developer.allizom.org
- stage.mdn.moz.works
- stage-files.mdn.moz.works

I don't think the `arn:aws:acm:us-west-2:236517346949:certificate/536f299c-ebcb-457f-9ce2-6c98a5132c82` cert is valid any more.

escattone: Ah, thanks @metadave! I think what happened is that I had a cert ARN from ages ago, and since then you've manually added the cert above to the stage ELB for the `web` service. I'll fix the PR to use:

`arn:aws:acm:us-west-2:236517346949:certificate/ddf39755-b0e2-4bfa-92e5-7f2d6025db14`

escattone:
Good catch all! I updated the stage configuration scripts to reflect the actual value used by the stage ELB.

PR #618, opened 2017-10-24
Change default Deis RDS instance size to t2.small

jgmize:


jgmize:
@metadave as discussed, I've resized the existing instances in Tokyo & Frankfurt as well.

Issue #590, opened 2017-10-12
Add CDN for  attachments domain


metadave:
The attachments domain (currently mdn.mozillademos.org) serves four types of requests:

* ``/files`` like [files/12984/web-font-example.png](https://mdn.mozillademos.org/files/12984/web-font-example.png), which are user attachments. It is possible to replace a user attachment with a new file, and we can cache on the modification date.
* ``@api`` files like [@api/deki/files/3613/=hut.jpg](https://mdn.mozillademos.org/@api/deki/files/3613/=hut.jpg), which redirect to a ``/files`` endpoint. These could be cached for a long time.
* ``$samples`` like [/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color](https://mdn.mozillademos.org/en-US/docs/Learn/CSS/Styling_text/Fundamentals$samples/Color), which are extracted from user-written wiki content.  The content changes when the page is updated.
* 403 (Permission Denied), 404 (Not Found), and 50x (Server Error) pages

These are good candidates for a CDN. Unlike the static assets, we would want short cache times, so that if a file is removed or updated, it will take effect quickly.  We could also use the CloudFront Invalidation API.

Cache headers for ``/files`` were improved in #578.

NOTE: notes above by @jwhitlock, my name shows as I converted it from a project card to an issue.

jwhitlock:
We can base it off of:

* [The current prod CDN config](https://github.com/mozmeao/infra/blob/master/apps/mdn/mdn-aws/infra/mdn-cdn/mdn_cdn.tf#L24-L32)
* [Terraform settings](https://github.com/mozmeao/infra/blob/master/apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront/cloudfront.tf)

Possibly relevant CloudFront docs:

* [Values that you specify when you create or update a web distribution](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html) - settings we can set
  * [Viewer Protocol Policy](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesViewerProtocolPolicy) - Redirect HTTP to HTTPS
  * [Allowed HTTP Methods](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesAllowedHTTPMethods) - GET, HEAD, OPTIONS
[...]

PR #622, opened 2017-10-25
add mdn attachments cdn

metadave:
See https://github.com/mozmeao/infra/issues/590

### DIFF
--- a/apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_attachments/cloudfront_attachments.tf
+++ b/apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_attachments/cloudfront_attachments.tf
@@ -18,7 +18,6 @@ resource "aws_cloudfront_distribution" "mdn-attachments-cf-dist" {
   custom_error_response {
     error_caching_min_ttl = 300
     error_code            = 404
-    response_page_path    = "/404.html"
   }
 
   default_cache_behavior {
--- a/apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_attachments/cloudfront_attachments.tf
+++ b/apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_attachments/cloudfront_attachments.tf
@@ -1,0 +1,73 @@
+variable "distribution_name" {}
+variable "comment" {}
+variable "domain_name" {}
+variable "acm_cert_arn" {}
+
+variable "aliases" {
+  type = "list"
+}
+
+resource "aws_cloudfront_distribution" "mdn-attachments-cf-dist" {
+  aliases         = "${var.aliases}"
+  comment         = "${var.comment}"
+  enabled         = true
+  http_version    = "http2"
+  is_ipv6_enabled = false
+  price_class     = "PriceClass_All"
+
+  custom_error_response {
+    error_caching_min_ttl = 300
+    error_code            = 404
+    response_page_path    = "/404.html"
+  }
+
+  default_cache_behavior {
+    allowed_methods = ["GET", "HEAD", "OPTIONS"]
+    cached_methods  = ["GET", "HEAD", "OPTIONS"]
+    compress        = true
+
+    default_ttl     = 86400
+    max_ttl         = 31536000
+    min_ttl         = 0
+
+    smooth_streaming       = false
+    target_origin_id       = "${var.distribution_name}"
+    viewer_protocol_policy = "redirect-to-https"
+
+    forwarded_values {
+      query_string = false
+
+      cookies {
+        forward = "none"
+      }
+    }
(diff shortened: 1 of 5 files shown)


metadave:
Please verify these values:

- `response_page_path`
- `default_ttl`
- `max_ttl`

apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_attachments/cloudfront_attachments.tf
```diff
@@ -0,0 +1,73 @@
...
+  comment         = "${var.comment}"
+  enabled         = true
+  http_version    = "http2"
+  is_ipv6_enabled = false
+  price_class     = "PriceClass_All"
+
+  custom_error_response {
+    error_caching_min_ttl = 300
+    error_code            = 404
+    response_page_path    = "/404.html"
```
jwhitlock: This page doesn't exist:

https://mdn.mozillademos.org/404.html

I think the line should be removed.

apps/mdn/mdn-aws/infra/mdn-cdn/mdn_cdn.tf
```diff
@@ -32,3 +33,15 @@ module "mdn-cloudfront-prod" {
 }
 
 
+# Attachments origin
+
+module "mdn-cloudfront-attachments-prod" {
+    source = "./cloudfront_attachments"
```
jwhitlock: I guess this is why the terraform file has to be in a subdirectory

apps/mdn/mdn-aws/infra/mdn-cdn/mdn_cdn.tf
```diff
@@ -32,3 +33,15 @@ module "mdn-cloudfront-prod" {
...
+
+    acm_cert_arn = "arn:aws:acm:us-east-1:236517346949:certificate/4322464c-b6d9-44d9-914d-3554461bcbb3"
```
jwhitlock: ``us-east-1``? Is this right?

metadave: all cloudfront certs have to be provisioned in us-east-1.

apps/mdn/mdn-aws/infra/mdn-cdn/mdn_cdn.tf
metadave: yes, it's just the subdir name for additional terraform files. It helps with Terraform reuse.

apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_attachments/cloudfront_attachments.tf
metadave: :+1:

jwhitlock:
Oops, forgot to look at ttl values. Reading the docs...

apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_attachments/cloudfront_attachments.tf
```diff
@@ -0,0 +1,72 @@
...
+    error_code            = 404
+  }
+
+  default_cache_behavior {
+    allowed_methods = ["GET", "HEAD", "OPTIONS"]
+    cached_methods  = ["GET", "HEAD", "OPTIONS"]
+    compress        = true
+
+    default_ttl     = 86400
+    max_ttl         = 31536000
```
jwhitlock: In https://github.com/mozilla/kuma/pull/4476, we're setting the ``Cache-Control`` headers for most responses, so [according to the docs](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html) we'll get the lesser of ``max_ttl`` and the ``max-age`` we're sending.  For some we're sending 5 minutes (60 * 5 = 300), others 5 days (60 * 60 * 24 * 5 = 432000). I think we're not sending caching headers on some error responses.

I'm more comfortable with 24 hours for default (86400, current setting), and 5 days for max (432000). 1 year seems like asking for trouble for the next devs.

metadave: fixed in the latest push.

btw, the 1 year value came from the SCL3 Cloudfront config :-)

jwhitlock: 😳

metadave:
Certs have already been updated on prod ELBs in Portland && Frankfurt

metadave:
terraform has been applied

PR #623, opened 2017-10-25
add kuma ATTACHMENT_ORIGIN setting

escattone:
Companion PR to https://github.com/mozilla/kuma/pull/4479.


apps/mdn/mdn-aws/k8s/kuma.base.yaml.j2
```diff
@@ -33,6 +33,8 @@
               value: "{{ KUMA_ALLOWED_HOSTS }}"
             - name: ATTACHMENT_HOST
               value: {{ KUMA_ATTACHMENT_HOST }}
+            - name: ATTACHMENT_ORIGIN
+              value: {{ KUMA_ATTACHMENT_ORIGIN | default(KUMA_ATTACHMENT_HOST) }}
```
jwhitlock: Let's use:

```
value: {{ KUMA_ATTACHMENT_ORIGIN | default(KUMA_ATTACHMENT_HOST, True) }}
```

This way, [a blank string](http://jinja.pocoo.org/docs/2.9/templates/#default) also defaults to ``KUMA_ATTACHMENT_HOST``

PR #624, opened 2017-10-25
change mdn cdn origin to d.m.o

metadave:


metadave:
tf applied

metadave:
paired with @jgmize on this

PR #625, opened 2017-10-25
use $HOME instead of ~ to satisfy jenkins

escattone:
After banging my head against the wall of Jenkins for too long, @metadave saved me from further anguish by informing me that Jenkins doesn't like `~`. Thanks @metadave!


metadave:
:+1:

PR #632, opened 2017-10-31
improve apps/mdn/mdn-aws/k8s/README.md

escattone:
https://github.com/mozmeao/infra/pull/629 and https://github.com/mozilla/kuma/pull/4484 reminded me that the current documentation on deploying MDN did not yet include database migration. This PR remedies that and makes the documentation clearer I think.


metadave:
:+1:

PR #638, opened 2017-11-06
SUMO infra rough draft

metadave:
probably a bit premature, but here's Terraform that we'll use for SUMO infra provisioning. It hasn't been applied (no AWS resources have been provisioned), but I'll be iterating on this in the future.


metadave:
updated with backup/maintenance windows as specified in https://github.com/mozilla/kitsune/issues/2942#issuecomment-342504999

PR #656, opened 2017-11-28
Add ENABLE_CANDIDATE_LANGUAGES=True for stage, dev

jwhitlock:
This enables candidate languages on the staging server, so that the localization team can see UI strings in context. See bug 984149.


metadave:
👍

PR #657, opened 2017-11-28
updates for sumo dev rds provisioning

metadave:


apps/sumo/infra/multi_region/tf/main.tf
```diff
@@ -67,9 +84,10 @@ module "mysql-prod" {
     mysql_instance_class = "db.m4.xlarge"
```
jgmize: Let's start with a t2.small for dev

metadave: [dev is already a t2.small](https://github.com/mozmeao/infra/pull/657/files/1b7a1af824d1bfd37c95b33c711f244fc9c11410#diff-62417988467dd21bf065e0822b93f50eR49), the prod section is commented out for now.

PR #659, opened 2017-11-29
Add Kuma env for FOUNDATION_CALLOUT

jwhitlock:
When enabled, this adds a call-to-action on the homepage to donate to the Mozilla Foundation. It goes up from late November to the first week of January. It depends on mozilla/kuma#4572 for 2017.


jwhitlock:
Tested with stage deploy (after a rebase, since my master was _way_ out of date)

PR #663, opened 2017-11-30
don't perform service changes unless CHANGE_MDN_INFRA is defined

metadave:
this change checks to see if the environment variable `CHANGE_MDN_INFRA` is set (to any value) before performing any infra related tasks (or things that probably shouldn't be deleted). It's best ***NOT*** to export `CHANGE_MDN_INFRA`.

Examples:

Without setting the env var:
```
$ make k8s-services
Makefile:518: *** CHANGE_MDN_INFRA is undefined, set it to any value to allow infra changes.  Stop.
```

Setting the env var and calling an infra target:
```
CHANGE_MDN_INFRA=true make k8s-services
# make target runs!
```


metadave:
@jwhitlock do you have any opinion on this change?

PR #664, opened 2017-11-30
add infra tests for MDN ELBs

metadave:
here's an experiment. 

This PR has some pytest code that tests stage/prod ELBs to ensure they're configured correctly, specifically:
- is the ELB configured to use port 80, and is traffic sent to the redirector NodePort?
- is the ELB configured to use port 443, and is traffic sent to the mdn web service NodePort?
- are ELB listener protocols specified correctly?
- is ELB logging enabled?
- is the connection timeout configured to use the same value as specified in the `apps/mdn/mdn-aws/k8s/Makefile`?

These tests depend on values specified in the region env files located in `apps/mdn/mdn-aws/k8s/regions/`. I'm using `os.environ` without checking for empty values: if they're empty, the test _should_ fail. Also, some boto3/k8s queries blindly access array elements, if they don't work... something is wrong with the env.

(autopep8 made my long lines ugly.)

***Do NOT run this in Jenkins or any tool that will expose build logs as failure output includes security creds etc.***

### DIFF
--- a/apps/mdn/mdn-aws/infra_testing/tests/test_listeners.py
+++ b/apps/mdn/mdn-aws/infra_testing/tests/test_listeners.py
@@ -1,0 +1,139 @@
+# -*- coding: utf-8 -*-
+
+import boto3
+from kubernetes import client, config
+import os
+import pytest
+
+# fixtures starting with k8s_ return K8s API objects
+#   documented here: https://github.com/kubernetes-incubator/client-python/blob/master/kubernetes/README.md
+# fixtures starting with aws_ return boto3 objects
+#   documented here: http://boto3.readthedocs.io/en/latest/index.html
+
+
+@pytest.fixture(scope="module")
+def k8s():
+    """A Kubernetes client"""
+    config.load_kube_config(os.environ['KUBECONFIG'])
+    return client.CoreV1Api()
+
+
+def k8s_web_service():
+    """Returns the MDN web service K8s object"""
+    services = k8s().list_namespaced_service(os.environ['K8S_NAMESPACE'])
+    web_service = [
+        elb for elb in services.items if elb.metadata.name == 'web'][0]
+    return web_service
+
+
+@pytest.fixture(scope="module")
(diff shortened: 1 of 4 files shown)


apps/mdn/mdn-aws/infra_testing/tests/test_listeners.py
```diff
@@ -0,0 +1,139 @@
...
+def aws_web_service_elb():
+    elb_host = k8s_web_service_elb_hostname()
+    result = aws_elb_client().describe_load_balancers(
+        LoadBalancerNames=[elb_host])
+    return result['LoadBalancerDescriptions'][0]
+
+
+class TestELB(object):
+    def test_k8s_connection(self):
+        assert k8s_web_service_elb_hostname() is not None
```
jgmize: NIT: `is not None` is unecessary

apps/mdn/mdn-aws/infra_testing/tests/test_listeners.py
```diff
@@ -0,0 +1,139 @@
...
+# fixtures starting with k8s_ return K8s API objects
+#   documented here: https://github.com/kubernetes-incubator/client-python/blob/master/kubernetes/README.md
+# fixtures starting with aws_ return boto3 objects
+#   documented here: http://boto3.readthedocs.io/en/latest/index.html
+
+
+@pytest.fixture(scope="module")
+def k8s():
+    """A Kubernetes client"""
+    config.load_kube_config(os.environ['KUBECONFIG'])
```
jgmize: I don't normally set a KUBECONFIG env var and use the default file location instead. How about we use `os.environ.get('KUBECONFIG')` here instead?

metadave: 👍

apps/mdn/mdn-aws/infra_testing/tests/test_listeners.py
metadave: fixed

Issue #660, opened 2017-11-29
Eliminate DB dependency from liveness & readiness endpoints in Kuma


jgmize:
When MySQL is slow to respond, these endpoints are slow to respond, which causes a cascading failure situation.

Original implementation in https://github.com/mozilla/kuma/pull/4240

escattone:
Thanks for bringing this up @jgmize. In my original middleware-based implementation, the liveness endpoint did not depend on MySQL, just the readiness endpoint. In the current view-based implementation, since requests go through the entire middleware stack (which hits MySQL), now both endpoints dependent on MySQL.

I think we should at least move the liveness endpoint back to a middleware implementation situated at the very beginning of the stack, so it no longer depends on MySQL. It seems to me that the readiness endpoint should remain dependent on MySQL, since Kuma isn't ready without it, but maybe we need to increase the K8s timeout for readiness probes?

jgmize:
There was another downtime event this morning where a scraper was hitting a bunch of URLs with poorly optimized SQL queries, resulting in increased DB latency and therefore healcheck and liveness probe failures. I temporarily removed the probes specified in this template https://github.com/mozmeao/infra/blob/master/apps/mdn/mdn-aws/k8s/kuma.deploy.yaml.j2#L15-L32 via `kubectl edit`.

jgmize:
> It seems to me that the readiness endpoint should remain dependent on MySQL, since Kuma isn't ready without it, but maybe we need to increase the K8s timeout for readiness probes?

No, please remove the DB dependency from both probes. Leaving the dependency in the readiness probes leaves us open to another cascading failure situation.

metadave:
Whats the best way to monitor MySQL perf if the probes are removed?

jwhitlock:
If web pods are unable to connect to the database, we'll see a significant increase in 500 errors.  New Relic will alert us at 5% or greater. If the database connection fails, the homepage may also fail, and we'll get a downtime alert.

Here's the average response time during the event in New Relic. This gives a visual indication that something is up:

<img width="635" alt="nr response time" src="https://user-images.githubusercontent.com/286017/33436619-325df3c6-d5ab-11e7-9b70-b9f42b74575f.png">

Here's Database connections. It seems like a weaker signal, but does show a spike during the incident. I suspect that if Kubernetes wasn't killing web pods, a sustained high # of DB connections would be a good signal, but it would be proportionate to the total number of pods, so the math would be hard for alerting:

[...]

PR #665, opened 2017-11-30
increase threshold of failure for liveness/readiness probes

escattone:
After a [lengthy discussion](https://github.com/mozmeao/infra/issues/660), we decided to increase the threshold of failure for the liveness and readiness probes for both kuma and kumascript. In general, we want to be much more certain that re-routing traffic (readiness failure) or restarting (liveness failure) will lead to a successful outcome for the system.


metadave:
thanks @escattone!

PR #670, opened 2017-12-04
Add log parser for MDN papertrail logs

jwhitlock:
This is built to be a generic log parser, but doesn't do a very good job yet. It currently is used to find the top IPs in the logs from a downtime incident, to help determine if an IP ban would be useful, or for the downtime post-mortem.


metadave:
👏 👏 👏

PR #674, opened 2017-12-06
Add RATELIMIT_ENABLE, default true for web, false for api

jgmize:


apps/mdn/mdn-aws/k8s/Makefile
```diff
@@ -211,6 +211,9 @@ export DATADOG_REDIS_CONFIG ?= foo
 # see https://kubernetes.io/docs/concepts/configuration/secret/
 export DATADOG_REDIS_CONFIG_BASE64 ?= $(shell echo -n "${DATADOG_REDIS_CONFIG}" | base64)
 
+export API_RATELIMIT_ENABLE ?= false
+export WEB_RATELIMIT_ENABLE ?= true
```
escattone: A tiny nit, but would you mind moving each of these up into their respective groups?

escattone:
I took the liberty of taking care of my "nit", and then realized that we need to set `KUMA_RATELIMIT_ENABLE` for the Celery deployments, so I did that as well.

PR #676, opened 2017-12-07
only check CHANGE_MDN_INFRA for check-service-env target

escattone:
When `CHANGE_MDN_INFRA` is not set, **all** targets fail since [this](https://github.com/mozmeao/infra/blob/c9c528919ad739d623b60b74bb898293e9ca029c/apps/mdn/mdn-aws/k8s/Makefile#L524-L526) is always run to determine what the Makefile actually "sees". I created a new bash script `check_infra_lock.sh` which is called by the `check-service-env` target, so `CHANGE_MDN_INFRA` is only checked when the target is run.


metadave:
ouch! Sorry about that. 
+1!

escattone:
No worries @metadave, thanks so much for the quick review!

PR #680, opened 2018-01-03
parse_log.py: Suggest the last 30 minutes of logs

jwhitlock:
In the usage for MDN's ``parse_log.py`` helper, suggest downloading the most recent 30 minutes of logs from Papertrail.


metadave:
:+1:

PR #681, opened 2018-01-03
Turn off Mozilla Foundation fundraiser

jwhitlock:
The end of year Mozilla Foundation fundraiser is over, so turning it off.


jwhitlock:
Maybe we can find a way to reduce duplication in these files in the new year. Until then, here's a multi-file search and replace that worked for me (probably macOS specific):

```
cd apps/mdn/mdn-aws/k8s/regions
find . -name '*.sh' -exec sed -i "" -e 's/KUMA_FOUNDATION_CALLOUT=True/KUMA_FOUNDATION_CALLOUT=False/g' {} \;
```

metadave:
:+1:

PR #682, opened 2018-01-04
Use Redis not MySQL as the Celery results backend

jwhitlock:
Building on mozilla/kuma#4615, use Redis DB 1 as the Celery results backend, to reduce reads/writes against MySQL.


metadave:
looks good from the infra POV

PR #687, opened 2018-01-10
Incident response guide rough draft

metadave:
cc @jgmize @bensternthal @pmac @jwhitlock @escattone

### DIFF
--- a/docs/incident_response.md
+++ b/docs/incident_response.md
@@ -18,7 +18,7 @@ When an incident occurs, the IL posts messages in the primary developer IRC chan
 - finally, decide if you need to notify the appropriate engineers that can help resolve the issue
     - TODO: link to on-call process
 
-These channels are:
+Our primary service channels are:
 - `#mdndev`
 - `#www`
 - `#basket`
--- a/docs/incident_response.md
+++ b/docs/incident_response.md
@@ -1,0 +1,104 @@
+# MozMEAO incident response guide
+
+Stay cool, take a deep breath. We'll solve this problem together, and this guide is here to help.
+
+This document does not describe the on-call or escalation process, but rather the process we use once an issue has been identified.
+
+## Declaring an incident and the incident leader (IL) role
+
+The incident leader (IL) serves as the primary point of contact and coordinator for an incident. This person may have been called by the moc, or just noticed that a service is broken. 
+
+When an incident occurs, the IL posts messages in IRC stating:
+
+- an incident has been declared and a few words about the incident
+- "I am the incident leader"
+- the main IRC channel for communication
+    - we'll call this the *incident comms channel* in this document
+- a Vidyo chat room if needed
+- finally, decide if you need to notify the appropriate engineers that can help resolve the issue
+    - TODO: link to on-call process
+
+***Security incidents must not be discussed on IRC.***
+
+Example incident declaration:
+
+```
+metadave>  MDN is experiencing an outage                          12:00 PM
+metadave>  I am declaring an incident and am incident leader      12:00 PM
+metadave>  updates will be posted in the #mdndev channel          12:00 PM
+metadave>  ping jwhitlock rjohnson there is an urgent MDN outage  12:00 PM
+metadave>  please meet in my Vidyo room                           12:01 PM
+```
+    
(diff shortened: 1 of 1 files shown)


docs/incident_response.md
```diff
@@ -0,0 +1,106 @@
...
+
+Stay cool, take a deep breath. We'll solve this problem together, and this guide is here to help.
+
+This document does not describe the on-call or escalation process, but rather the process we use once an issue has been identified.
+
+## Declaring an incident and the incident leader (IL) role
+
+The incident leader (IL) serves as the primary point of contact and coordinator for an incident. This person may have been called by the moc, or just noticed that a service is broken. 
+
+When an incident occurs, the IL posts messages in IRC stating:
```
bensternthal: Which IRC channel is this information posted to?

docs/incident_response.md
```diff
@@ -0,0 +1,106 @@
...
+
+> don't post phone numbers in IRC!
+
+### Frequency of updates
+
+For outages, status should be posted every 30 minutes *or less* in the incident channel.
+
+## Incident timeline
+
+To help write an incident report, it's very useful to include a `TL:` prefix on IRC messages when logging important events and decisions.
```
bensternthal: 👍

escattone:
This looks good so far. The only question that arose for me was the same one that @bensternthal noted above, namely that it wasn't clear which IRC channel should be used for the initial incident announcement, or perhaps we simply choose the IRC channel based on the product, so `#mdndev` for MDN?

docs/incident_response.md
metadave: I was thinking that a message should be posted in the primary channel where backend engineers idle for each app:
- #mdndev
- #www
- #basket
- #snippets
- #sumo
etc.

metadave:
This doc was based on things I remember working well from my time at [Chef](https://www.chef.io/). We definitely don't have to adopt this process if it's too heavy etc. How do folks feel?

bensternthal:
I support this. Does not feel to heavy to me.

safwanrahman:
There should be some instruction about when to knock the developer.
Moreover, it would be better if all the components developer names are mentioned.

docs/incident_response.md
```diff
@@ -0,0 +1,106 @@
...
+    - the moc maintains an incident respone guide [here](https://mana.mozilla.org/wiki/display/SECURITY/Incident+Response#IncidentResponse-IncidentResponseTemplate).
+
+Example:
+
+```
+metadave>  there is an MDN outage, we are working in #mdndev       12:05 PM
+metadave>  I'll post an update here within the next 30 minutes     12:05 PM
+```
+
+- Notify team lead for the product impacted.
```
safwanrahman: Team leader means product manager or dev lead?

safwanrahman:
For a longer downtime, a bug should be filled describing the issue and ask `moc` to update the status page with the bug.

metadave:
> There should be some instruction about when to knock the developer.
> Moreover, it would be better if all the components developer names are mentioned.

@safwanrahman that is beyond the scope of this document, see the second line. Contact information is maintained in our internal wiki.

docs/incident_response.md
```diff
@@ -0,0 +1,118 @@
...
+- a Vidyo chat room if needed
+- finally, decide if you need to notify the appropriate engineers that can help resolve the issue
+    - TODO: link to on-call process
+
+Our primary service channels are:
+- `#mdndev`
+- `#www`
+- `#basket`
+- `#snippets`
+- `#sumo`
```
safwanrahman: I think it should be `#sumodev`

docs/incident_response.md
```diff
@@ -0,0 +1,118 @@
+# MozMEAO incident response guide
+
+Stay cool, take a deep breath. We'll solve this problem together, and this guide is here to help.
+
+This document does not describe the on-call or escalation process, but rather the process we use once an issue has been identified.
+
+## Declaring an incident and the incident leader (IL) role
+
+The incident leader (IL) is a Mozillian on staff that serves as the primary point of contact and coordinator for an incident. This person may have been called by the moc, or just noticed that a service is broken.
```
safwanrahman: Mozillian or Staff?

metadave: Mozilla staff only, fixed.

escattone:
@metadave Thanks for this! I do not feel it's too heavy. I think it's worth adopting and we can adjust as we see fit.

Issue #693, opened 2018-01-16
Kitsune static file CDN provisioning


metadave:
https://github.com/mozilla/kitsune/issues/2949#issuecomment-357668175

- [x] request certificates
- [x] automate CDN provisioning
- [x] create route53 entries
- [x] request webops DNS change

> static-media-{dev, stage, prod}-cdn.sumo.mozilla.net CNAME to static-media-{dev,stage,prod}-cdn.sumo.moz.works

metadave:
https://github.com/mozmeao/infra/pull/694

PR #694, opened 2018-01-16
SUMO static media CDN

metadave:
@pmac @jgmize @glogiotatidis 

There are several TODOs. If the overall config is ok with you, I'll request ACM certs, dns changes from webops, and create the Route53 entries for .moz.works.

I was unsure what the `domain_name` should be for the dev instance. `domain_name` here is actually the name of the Cloudfront origin, but I'm not going to rename it or Terraform will flip out. Please double check that the stage/prod origins are the droids you're looking for.

See also:
- [Terraform docs for Cloudfront](https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html)
- https://github.com/mozmeao/infra/issues/693
- https://github.com/mozilla/kitsune/issues/2949#issuecomment-357668175

### DIFF
--- a/apps/sumo/infra/shared/main.tf
+++ b/apps/sumo/infra/shared/main.tf
@@ -96,7 +96,7 @@ module "sumo-static-media-dev-cf" {
 module "sumo-static-media-stage-cf" {
     source = "./cloudfront_static_media"
 
-    acm_cert_arn = " arn:aws:acm:us-east-1:236517346949:certificate/11d5784f-997e-4271-a6d5-2cfeb8dccb27"
+    acm_cert_arn = "arn:aws:acm:us-east-1:236517346949:certificate/11d5784f-997e-4271-a6d5-2cfeb8dccb27"
     aliases = ["static-media-stage-cdn.sumo.mozilla.net", "static-media-stage-cdn.sumo.moz.works"]
     comment = "Stage CDN for SUMO static media"
     distribution_name = "SUMOStaticMediaStageCDN"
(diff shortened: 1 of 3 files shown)


metadave:
I'm going to apply the [whitenoise changes](http://whitenoise.evans.io/en/stable/django.html#restricting-cloudfront-to-static-files) first, and then I'll reopen.

apps/sumo/infra/shared/main.tf
```diff
@@ -75,3 +80,35 @@ module "sumo-user-media-prod-cf" {
...
+# static media CDN
+#####################################################################
+module "sumo-static-media-dev-cf" {
+    source = "./cloudfront_static_media"
+
+    acm_cert_arn = "TODO"
+    aliases = ["static-media-dev-cdn.sumo.mozilla.net", "static-media-dev-cdn.sumo.moz.works"]
+    comment = "Dev CDN for SUMO static media"
+    distribution_name = "SUMOStaticMediaDevCDN"
+    domain_name = "TODO"
```
metadave: sumo-dev.frankfurt.moz.works

metadave: - dev.sumo.moz.works CNAME -> ELN (https) -> sumo-dev.frankfurt.moz.works
- we will need a valid cert

metadave:
note to @metadave: request certs in `us-east-1` for Cloudfront

metadave:
Ok, certs requested, dev name updated. 

DNS changes requested here: https://bugzilla.mozilla.org/show_bug.cgi?id=1433561

metadave:
Cloudfront provisioned, but won't work until we get backend services up and running with correct certs.

PR #711, opened 2018-01-24
Base mdn-backup-cron image on python:3.6-slim

jwhitlock:
We're moving away from ``mozmar/base``, and toward language-specific base images.

This may also help an issue where the Kuma images (using ``python:2.7-slim``) are using ``LANG=C.UTF-8``, and the ``quay.io/mozmar/base:latest`` image is leaving ``LANG`` unset and defaulting to ``LC_CTYPE=POSIX``. This may be the reason that Kuma images are creating filenames with unicode characters that the backup process can't read.

### DIFF
--- a/apps/mdn/utils/mdn_backup_cron/image/Dockerfile
+++ b/apps/mdn/utils/mdn_backup_cron/image/Dockerfile
@@ -1,4 +1,4 @@
-from quay.io/mozmar/base:latest
+from python:3.6-slim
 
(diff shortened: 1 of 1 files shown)


metadave:
for @metadave https://bugzilla.mozilla.org/show_bug.cgi?id=1271743

metadave:
This is probably fine, but we'll need to update the `Makefile` with new image version and test a few runs of the script.

PR #713, opened 2018-01-26
Initial SUMO app automation

metadave:
This will be a work-in-progress until some time before the migration. We'll be adding new targets, tweaking environment vars. It's the same type of [automation that's used for MDN](https://github.com/mozmeao/infra/tree/master/apps/mdn/mdn-aws/k8s): it's easy to work with and there are no surprises (you can render yaml right to the terminal instead of sending to k8s etc).

When we're ready to test this, I'll create a secrets resource (a tweaked version of the one that Deis created, but using the template in `apps/sumo/k8s/sumo-secrets.yaml`). We may also need to work out some service definitions.


metadave:
I'm thinking of switching the Makefile to http://www.pyinvoke.org/, but I need to think about how to structure config files.

metadave:
/me is working on a pyinvoke-based deployment

metadave:
https://github.com/mozmeao/infra/pull/718

PR #714, opened 2018-01-26
Python-based ELB provisioning tool

metadave:
It's alive.

See README.md for more info, and the first Oregon-b ELB is defined in `elb_automation/meaoelb/oregon_b.py`


elb_automation/meaoelb/config.py
```diff
@@ -0,0 +1,153 @@
+import pprint
+
+
+class DictLike(dict):
```
jgmize: FYI Python 3.7 data classes look like they might be good for this use case in the future: https://www.python.org/dev/peps/pep-0557/

elb_automation/meaoelb/config.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def aws_merge(self, d):
+        d['ConnectionSettings'] = {'IdleTimeout': self.timeout}
+
+
+class ELBAtts():
+    def __init__(self, *args):
+        self.items = []
+        for arg in args:
+            self.items.append(arg)
```
jgmize: NIT: 
```python
self.items = list(args)
```

elb_automation/meaoelb/elb_ctx.py
```diff
@@ -0,0 +1,173 @@
...
+            return
+
+        if self.dry_run_mode:
+            print(
+                "\t➤ ELB {} would have been created".format(
+                    service_config.elb_config.name))
+            return
+
+        self._create_elb(service_config)
+        self._modify_elb_atts(service_config)
```
jgmize: Why do we create then immediately modify the elb, instead of creating the ELB with the desired attributes? Is this working around some limitation of the AWS api?

metadave: great question - it's because the [create_load_balancer](https://boto3.readthedocs.io/en/latest/reference/services/elb.html#ElasticLoadBalancing.Client.create_load_balancer) call doesn't take any attributes :-(

jgmize: Answering my own question: I see from https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_CreateLoadBalancer.html that there doesn't appear to be a way to pass in https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_LoadBalancerAttributes.html so we have to use https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_ModifyLoadBalancerAttributes.html instead.

metadave:
I'd love to have automatic testing of each defined ELB, similar to [these tests](https://github.com/mozmeao/infra/tree/master/apps/mdn/mdn-aws/infra_testing)

metadave:
I'm going to followup with a PR that does ELB testing for us.

PR #718, opened 2018-01-29
 PyInvoke-based SUMO deployment 

metadave:
- includes K8s "vanilla" deployment templates


pmac:
Digging the `invoke` usage. Looks like it might work out well. I'm wondering though whether much of this doesn't belong in the kitsune repo instead, and the `deploy_utils.py` bits in an installable helper lib? Is that the long-term plan?

metadave:
>  I'm wondering though whether much of this doesn't belong in the kitsune repo instead, and the deploy_utils.py bits in an installable helper lib? Is that the long-term plan?

Whatever works for ya'll, we kept the MDN deployment stuff in this repo for now, but it may be desirable to move to Kitsune.

metadave:
moved to https://github.com/mozilla/kitsune/pull/3034

PR #723, opened 2018-01-31
bedrock-dev elb uses 80+443 listeners to communicate with nodeport

metadave:
Redirector service is not used in this config


metadave:
paired with @jgmize on this, so merging

PR #736, opened 2018-02-15
fix the MDN Frankfurt prod.mm.sh KUBECONFIG path for Jenkins

escattone:
This fix enables `kubectl` to be used within Jenkins for deploying to the Frankfurt K8s cluster.

Thanks to @metadave and @jwhitlock for all of their help with this!


jwhitlock:
We should fix or delete apps/mdn/mdn-aws/k8s/regions/portland/dev.sh as well.

PR #746, opened 2018-03-05
change SUMO stage/prod CDN to use traffic policies

metadave:


metadave:
paired with @jgmize on this one, going to merge.

PR #747, opened 2018-03-07
SUMO failover CDN

metadave:
We'll need a us-east-1 cert with `support.mozilla.org`, `support.mozilla.com`, `failover-cdn.sumo.moz.works`


metadave:
@glogiotatidis any thoughts on the `default_ttl`, `min_ttl` and `max_ttl` values?

apps/sumo/infra/shared/main.tf
```diff
@@ -112,3 +112,19 @@ module "sumo-static-media-prod-cf" {
...
     domain_name = "prod-tp.sumo.moz.works"
 }
+
+#####################################################################
+# failover CDN
+#####################################################################
+module "sumo-failover-cf" {
+    source = "./cloudfront_failover"
+
+    acm_cert_arn = "TODO"
```
jgmize: arn:aws:acm:us-east-1:236517346949:certificate/c677a888-a2ee-4486-92f5-cc99355624b7

metadave: PR updated with ARN

glogiotatidis:
> @glogiotatidis any thoughts on the default_ttl, min_ttl and max_ttl values?

yes. If I understand correctly settings `min` and `max` will override any headers send by the origin. We don't want that, as we'll be setting cache headers on the Django level.  I suggest that we set the default to something less aggressive like 4 hours and leave `min` and `max` unset so the app can override.

For reference the current default cache ttl for the Django App (when the cache setting is enabled) is 2 hours and it's configurable using an env variable. https://github.com/mozilla/kitsune/blob/master/kitsune/settings.py#L93-L95

metadave:
`min_ttl` and `max_ttl` are required values, I can try to set them to 0 and see what happens.

```
max_ttl (Required) - The maximum amount of time (in seconds) that an object is in a CloudFront cache before CloudFront forwards another request to your origin to determine whether the object has been updated. Only effective in the presence of Cache-Control max-age, Cache-Control 
s-maxage, and Expires headers.

min_ttl (Required) - The minimum amount of time that you want objects to stay in CloudFront caches before CloudFront queries your origin to see whether the object has been updated.
```

https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html

glogiotatidis:
Is this still open so we can configure the Vary headers?

metadave:
yes

glogiotatidis:
```
seadog@areti:~$ curl -I https://dev.sumo.moz.works
HTTP/1.1 302 FOUND
Cache-Control: no-cache, must-revalidate
Content-Type: text/html; charset=utf-8
Date: Tue, 13 Mar 2018 14:51:22 GMT
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Location: https://dev.sumo.moz.works/en-US/
Pragma: no-cache
Server: meinheld/0.6.1
Vary: Accept-Language, X-Mobile, User-Agent
X-Backend-Server: sumo-dev-web-4189611774-3xcj5.frankfurt.moz.works
x-content-type-options: nosniff
X-Frame-Options: DENY
X-Robots-Tag: noodp
x-xss-protection: 1; mode=block
Connection: keep-alive
```

glogiotatidis:
```
seadog@areti:~$ curl -I https://dev.sumo.moz.works/en-US/
HTTP/1.1 200 OK
Cache-Control: no-cache, must-revalidate
Content-Type: text/html; charset=utf-8
Date: Tue, 13 Mar 2018 14:50:47 GMT
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Pragma: no-cache
Server: meinheld/0.6.1
Vary: X-Mobile, User-Agent
X-Backend-Server: sumo-dev-web-4189611774-3xcj5.frankfurt.moz.works
x-content-type-options: nosniff
X-Frame-Options: DENY
X-Robots-Tag: noodp
x-xss-protection: 1; mode=block
Connection: keep-alive
```

jgmize:
We'll need to set the Host header to support.mozilla.org so that the l10n redirects work correctly: https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html#custom_header

metadave:
We can't set `Host`: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/forward-custom-headers.html

jgmize:
> We can't set Host: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/forward-custom-headers.html

That's unfortunate, because we can't set the origin to support.mozilla.org either. We'd have to change the l10n redirects at the app level then.

jgmize:
To clarify, the issue with the l10n redirects is that when you hit https://failover-cdn.sumo.moz.works it redirects to the CDN origin https://prod-frankfurt.sumo.moz.works/en-US/ instead of the host the client is using.

pmac:
We just have to set `USE_X_FORWARDED_HOST = True` and have the CDN send `X-Forwarded-Host: support.mozilla.org`.

jgmize:
https://failover-cdn.sumo.moz.works/en-US/ lgtm and l10n redirects send us to support.mozilla.org as intended. As discussed in IRC cloudfront is varying on both user-agent and accept-language on *all* urls in this configuration, not just those with the vary headers set, so the entire site will have a poor cache hit ratio until changes are made at the app level.

glogiotatidis:
> on both user-agent and accept-language 
and ftr x-mobile header

Thanks for working on this!

PR #757, opened 2018-03-27
bug 1431259: on prod, switch to primary CDN for static/media files

escattone:
This PR is blocked until the new primary CDN for production (`developer.mozilla.org`) has been created and is active. It switches the proxy for the static/media files from the asset CDN (`cdn.mdn.mozilla.net`) to the new primary CDN (`developer.mozilla.org`). It still allows requests from the the asset CDN, but they are permanently redirected to `developer.mozilla.org` (via the `LegacyDomainRedirectsMiddleware` middleware since `cdn.mdn.mozilla.net` is added to `LEGACY_HOSTS`).


jwhitlock:
I'd like to test on stage.mdn.moz.works for a few days before touching production.

PR #773, opened 2018-04-11
Add snippets-prod in oregon-b

jgmize:
Also remove unused bedrock-dev


elb_automation/meaoelb/oregon_b.py
```diff
@@ -57,14 +49,24 @@
 snippets_stage.elb_config.elb_atts.connection_settings.idle_timeout = 120
 snippets_stage.elb_config.health_check.target_path = '/healthz/'
 
+# ### Snippets Prod
+snippets_stage = elb_tool.define_elb_http(
+    service_namespace='snippets-prod',
+    service_name='snippets-nodeport',
+    ssl_arn='arn:aws:acm:us-west-2:236517346949:certificate/fbe34166-ae87-43f8-b9cc-7bc9a45d904c')
+snippets_stage.elb_config.elb_atts.connection_settings.idle_timeout = 120
```
glogiotatidis: We have `WSGI_KEEP_ALIVE=90` in snippets-prod config. Should we make those two values match?

metadave: https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-idle-timeout.html


        To ensure that the load balancer is responsible for closing the connections to your instance, make sure 
        that the value you set for the HTTP keep-alive time is greater than the idle timeout setting on your 
        load balancer.

elb_automation/meaoelb/sumo_oregon_b.py
```diff
@@ -0,0 +1,27 @@
...
+# ### SUMO Stage
+sumo_stage = elb_tool.define_elb_http(
+    service_namespace='sumo-stage',
+    service_name='sumo-nodeport',
+    ssl_arn='arn:aws:acm:us-west-2:236517346949:certificate/192b6409-996e-46ac-a3d9-c78a69670dae')
+sumo_stage.elb_config.health_check.target_path = '/healthz/'
+sumo_stage.elb_config.name = 'sumo-stage-b'
+
+# ### SUMO prod
+sumo_prod = elb_tool.define_elb_http(
```
metadave: this one is tricky because it's bound to 2 ASGs, and with the `define_elb_http` method we'll only be bound to 1.
I wonder if it's easier to skip the convenience methods and define the entire elb, following by a call to the asg bind code:

- https://github.com/mozmeao/infra/blob/master/elb_automation/meaoelb/ap_northeast_1.py#L97-L147
- https://github.com/mozmeao/infra/blob/master/elb_automation/meaoelb/ap_northeast_1.py#L524
- https://github.com/mozmeao/infra/blob/master/elb_automation/meaoelb/elb_ctx.py#L147

jgmize: This was for the oregon-b specific ELB; the ELB that was bound to 2 ASGs was created manually in the console.

Issue #775, opened 2018-04-11
MDN prod CDN go-live


metadave:
setup MDN production primary CDN using [this PR](https://github.com/mozmeao/infra/pull/726)

metadave:
## Release plan:

- provision prod CDN
- update the `mdn-prod.moz.works.` CNAME to point at the new CDN

## Emergency rollback:

- update the `mdn-prod.moz.works.` CNAME to point at `prod.mdn.moz.works.`

---

current state:

```
dig developer.mozilla.org

; <<>> DiG 9.10.3-P4-Ubuntu <<>> developer.mozilla.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 56050
;; flags: qr rd ra; QUERY: 1, ANSWER: 5, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;developer.mozilla.org.		IN	A

;; ANSWER SECTION:
developer.mozilla.org.	60	IN	CNAME	mdn-prod.moz.works.
mdn-prod.moz.works.	60	IN	CNAME	prod.mdn.moz.works.
prod.mdn.moz.works.	60	IN	CNAME	a8583e1be9a3711e7b1e70287e8e34e8-2077023531.us-west-2.elb.amazonaws.com.
a8583e1be9a3711e7b1e70287e8e34e8-2077023531.us-west-2.elb.amazonaws.com. 60 IN A 34.212.89.102
a8583e1be9a3711e7b1e70287e8e34e8-2077023531.us-west-2.elb.amazonaws.com. 60 IN A 52.24.17.241
```

metadave:
@escattone @jwhitlock let's capture any tech details here for the MDN CDN switch next week.

escattone:
Does this make sense as a checklist?

- [x] Successful testing on stage.
- [x] Create the new primary CDN ([via this Terraform](https://github.com/mozmeao/infra/pull/726/files#diff-790cd9740f7816b0c7811190c12f0a69)) with an origin of `prod.mdn.moz.works` (`d2jhm933xpgpf7.cloudfront.net`)
- [x] Manually order the behaviors in the new primary CDN
[...]

PR #788, opened 2018-04-17
bug 1431259: add asset cdn cname to primary cdn cnames

escattone:
This PR adds the domain name for the asset CDN to the primary CDN's list of CNAME's. This is to accommodate traffic from stale asset URL's after switching the asset traffic away from the asset CDN to the new primary CDN. This is to support @jwhitlock's strategy for turning off the production asset CDN (see https://github.com/mozmeao/infra/issues/775#issuecomment-381168302).


jwhitlock:
After some discussion about legacy MDN domains, I think we *don't* want to add cdn.mdn.mozilla.net as a CNAME. Instead, we'll watch for traffic to die down as asset requests move to developer.mozilla.org, and then let the web servers handle the redirects for any legacy requests to these assets URLs.

Issue #708, opened 2018-01-23
migrate careers to Oregon-B


metadave:
https://github.com/mozmeao/infra/issues/705

glogiotatidis:
- [ ] Scale cron process up in oregon-b after completing the migration

jgmize:
-   careers-stage already moved
-   [X] create careers-prod nodeport
    
        cd ~/src/mozilla/infra/apps/careers/k8s
        kubectl apply -f careers-prod-nodeport.yaml
-   [X] fix https redirects
    
        deis config:set SECURE_SSL_REDIRECT=True -a careers-prod
-   [ ] get cert
    -   valid for careers-prod.oregon-b.moz.works & careers.mozilla.org
-   [ ] create ELB
-   [ ] route53
    -   [ ] create healthcheck using /readiness endpoint
    -   [ ] traffic policy
-   [ ] close ticket

jgmize:
Migration complete.

PR #790, opened 2018-04-17
Add careers-prod oregon-b ELB for #708

jgmize:


jgmize:
The ELB created with this PR is now in production.

metadave:
:+1:

PR #794, opened 2018-04-18
bug 1431259: change path patterns of some behaviors to handle zones

escattone:
This PR updates the path patterns of 10 of the pass-through doc-related editing/administrative behaviors so that they can handle zoned documents.

### DIFF
--- a/apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_primary/cloudfront.tf
+++ b/apps/mdn/mdn-aws/infra/mdn-cdn/cloudfront_primary/cloudfront.tf
@@ -131,7 +131,7 @@ resource "aws_cloudfront_distribution" "mdn-primary-cf-dist" {
 
   # 4
   cache_behavior {
-    path_pattern = "*/docs/*$edit"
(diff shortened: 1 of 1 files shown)


metadave:
```
<!:!>@epsilon:mdn-cdn (fix-for-zones-1431259)$ terraform_0.11.7 plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

aws_cloudfront_distribution.mdn-attachments-cf-dist: Refreshing state... (ID: ...)
aws_cloudfront_distribution.mdn-primary-cf-dist: Refreshing state... (ID: ...)
aws_cloudfront_distribution.mdn-cf-dist: Refreshing state... (ID: ...)
aws_cloudfront_distribution.mdn-primary-cf-dist: Refreshing state... (ID: ...)

------------------------------------------------------------------------

No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.
```

Issue #701, opened 2018-01-19
migrate basket to Oregon-b


metadave:

jgmize:
- [X] basket-dev
- [X] basket-stage
- [X] basket-prod
- [X] basket-admin
- [X] basket-admin-stage
- [ ] add NR monitoring
- [ ] scale clock procs up after scaling down in usw
- [ ] scale up celery workers
- [ ] ELBs
- [ ] update admin redirect url in this and all reagions

PR #798, opened 2018-04-19
Add basket ELBs and nodeport services in oregon-b

jgmize:
for #701
- [x] nodeport services
  - [x] kubectl apply
- [x] make basket-oregon-b
- [x] make apply-basket-oregon-b


metadave:
:+1:

PR #803, opened 2018-04-23
block AWS cronjob

metadave:


k8s/tools/block-aws/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM python:2-slim
```
jgmize: Have you tried this with python 3?

metadave: it was supposed to be python3! Great catch!

metadave:
- https://github.com/kubernetes-client/python-base/blob/789de6a60dea9442ee47601a788d92c653ea69e9/config/incluster_config.py
- https://github.com/kubernetes-client/python-base/blob/789de6a60dea9442ee47601a788d92c653ea69e9/config/incluster_config.py

k8s/tools/block-aws/build.sh
```diff
@@ -0,0 +1,5 @@
+GIT_COMMIT_RAW=$(git rev-parse HEAD)
+GIT_COMMIT=${GIT_COMMIT_RAW:0:7}
```
jgmize: This can be simplified to
`GIT_COMMIT=$(git rev-parse --short HEAD)`

metadave:
TODO: add DMS + secret

PR #813, opened 2018-04-27
Add option to use kubectl to work around kubernetes-client/python#411 

jgmize:


jgmize:
Wrapping kubectl also works around https://github.com/kubernetes-client/python/issues/520 which may affect us when we upgrade to k8s 1.10.

PR #817, opened 2018-05-02
bug 1431259: update primary prod cdn cert and aliases

escattone:
This updates the cert and adds `cdn.mdn.mozilla.net` to the alternate domain names for the primary production CDN for MDN.


metadave:
it can take awhile for the `Pending Validation` status to change. It's valid as of this morning.

PR #819, opened 2018-05-03
bug 1431259: match cdn terraform to existing configuration

escattone:
This PR updates the terraform for the MDN CDN's to match the reality in AWS after retiring the old asset CDN.

We (@metadave, @jgmize, and myself) decided to send the `cdn.mdn.mozilla.net` traffic (static files) directly to the ELB via the `legacy.mdn.moz.works` CNAME (so `cdn.mdn.mozilla.net` --> `cdn.mdn.moz.works` --> `legacy.mdn.moz.works` --> ELB) instead of directly to `developer.mozilla.org`. We did this partly because CloudFront would not allow us to add `cdn.mdn.mozilla.net` to the list of alternate domain names on the primary production CDN (although we could have worked around that by first using `*.mdn.mozilla.net`, then changing back to `cdn.mdn.mozilla.net` after deleting the old asset CDN), but also because we thought it would be cleaner to keep that legacy domain out of the CDN (and so lump these requests into the legacy path and let the permanent redirects eventually reduce the traffic to zero or near zero).

This means that we did not update the cert on the primary production CDN to the new cert that contains `cdn.mdn.mozilla.net`, and we did not add `cdn.mdn.mozilla.net` to its list of alternate domain names.

We deleted the old asset CDN.


apps/mdn/mdn-aws/infra/mdn-cdn/mdn_cdn.tf
```diff
@@ -25,26 +25,13 @@ module "mdn-primary-cloudfront-stage" {
 
 module "mdn-primary-cloudfront-prod" {
    source = "./cloudfront_primary"
-   acm_cert_arn = "arn:aws:acm:us-east-1:236517346949:certificate/68395c25-b5d4-4875-b85a-8c89b80c4e01"
-   aliases = ["developer.mozilla.org", "cdn.mdn.mozilla.net"]
+   acm_cert_arn = "arn:aws:acm:us-east-1:236517346949:certificate/144c40ad-1a60-4865-a252-58ff23961787"
```
metadave: just dropping a note here to confirm that this is the us-east-1 cert _without_ `cdn.mdn.mozilla.net`

metadave:
thanks for the detailed write-up @escattone!

PR #831, opened 2018-05-16
protocol.mozilla.org hosting

metadave:


apps/protocol/setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+source ../bin/common.sh
+check_neres_env
+
+neres add-monitor "Viewsourceconf prod" \
```
pmac: `s/Viewsourceconf/Protocol/`

apps/protocol/setup.sh
```diff
@@ -0,0 +1,12 @@
...
+    https://viewsourceconf.org \
```
pmac: `protocol.mozilla.org`

PR #872, opened 2018-06-22
Move MDN k8s configuration to mdn/infra

jwhitlock:
mdn/infra is now used for deployments to MozMEAO's k8s clusters. Removing the files from here to reduce potential confusion.


jgmize:
thanks @jwhitlock!

PR #879, opened 2018-07-03
update bedrock-stage and bedrock-prod ELBs to use new *.frankfurt.moz.works cert

metadave:
***the changes have NOT been applied in AWS.***


metadave:
applied in prod

PR #946, opened 2018-09-26
update SUMO ELB's to support Windows XP

metadave:
- I typed these ciphers in by hand. 
- asg tweaks to match what's in AWS


metadave:
I'll need to add Frankfurt to this, most likely on Thursday AM.

metadave:
SUMO in Frankfurt is CDN-only, so changing adding the additional cipher to the ELB cipher suite won't be helpful.

PR #963, opened 2018-10-26
Ignore slack errors in the example script

pmac:
Not doing this will fail a build due to notification failure, which should just be ignored.


metadave:
:+1:

PR #998, opened 2019-01-09
[Fix #997] Careers Static Terraform

### LINKED ISSUE #997: S3 bucket automation for careers
See https://github.com/mozmeao/infra/pull/669 for example

glogiotatidis:
Creates
 - prod bucket
 - staging bucket
 - logs bucket
 - IAM user with S3 permissions to the buckets above
 - CF distribution for staging
 - CF distribution for prod

I need help getting certificates for the CF distributions. We have certificates for careers but they are both on us-west-2.

The above has been applied. You can test the staging CF dist at https://dmin398ukjj8z.cloudfront.net/

### DIFF
--- a/apps/careers/tf/careers.tf
+++ b/apps/careers/tf/careers.tf
@@ -10,59 +10,6 @@ terraform {
   }
 }
 
-resource "aws_iam_user" "careers" {
-  name = "careers-upload"
-}
-
-resource "aws_iam_user_policy" "careers-upload" {
-  name = "careers-upload"
-  user = "${aws_iam_user.careers.name}"
-
-  policy = <<EOF
-{
-    "Version": "2012-10-17",
-    "Statement": [
-        {
-            "Effect": "Allow",
-            "Action": [
-                "s3:ListBucket"
-            ],
-            "Resource": [
-                "arn:aws:s3:::mozilla-careers"
-            ]
-        },
-        {
-            "Effect": "Allow",
-            "Action": [
-                "s3:*"
-            ],
-            "Resource": [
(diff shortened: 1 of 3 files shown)


glogiotatidis:
Marked as WIP, to update with the needed certs.

apps/careers/tf/careers.tf
```diff
@@ -0,0 +1,283 @@
...
+
+terraform {
+  backend "s3" {
+    bucket = "mozilla-careers-s3-provisioning-tf-state"
+    key    = "tf-state"
+    region = "us-west-2"
+  }
+}
+
+resource "aws_iam_user" "careers" {
```
metadave: unless it needs to be public, we store all IAM TF [here](https://github.com/mozmeao/infra-private/tree/master/aws/iam_policies)

metadave:
@glogiotatidis what names would you like on the certs?

apps/careers/tf/careers.tf
glogiotatidis: Removed from here and opened a PR against private repo. https://github.com/mozmeao/infra-private/pull/153

Thanks!

glogiotatidis:
> @glogiotatidis what names would you like on the certs?

careers-prod.moz.works
careers.mozilla.org 

and

careers-stage.moz.works 
careers.allizom.org 

Thanks!

metadave:
@jgmize can you review the CloudFront `aliases` values?

glogiotatidis:
Updated with cert arns.

metadave:
@jgmize can you review the CloudFront aliases values?

PR #1057, opened 2019-03-22
Expire metadata after 7 days.

glogiotatidis:
An automated job will fetch metadata from our buckets daily and after that they
are no longer of use.

@metadave I haven't applied this one yet b/c the dir still contains references to tokyo and I wanted to check with you first.

Thanks!

### DIFF
--- a/apps/snippets/infra/multi_region/tf/storage/storage.tf
+++ b/apps/snippets/infra/multi_region/tf/storage/storage.tf
@@ -44,6 +44,16 @@ EOF
     }
   }
 
+  lifecycle_rule {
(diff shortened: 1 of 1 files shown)


metadave:
+1, was `terraform plan` happy with the change?

metadave:
I don't see the storage.tf file being referenced as a submodule from [here](https://github.com/mozmeao/infra/blob/master/apps/snippets/infra/multi_region/tf/snippets.tf).

glogiotatidis:
OK added the rule manually via the console.

PR #1095, opened 2019-04-30
Add child-src CSP header to awebpodcast to allow player embed.

jpetto:
We need to allow `https://heartthis.at` and `https://stats.heartthis.at` (and potentially other subdomians?) for the player `<iframe>` embed to work.

Will `*.heartthis.at` work for both cases above?

You can see the `<iframe>` code [here](https://github.com/mozilla/awebpodcast/blob/master/layouts/partials/episode-module.html#L43-L45).

Quick reference:

```
<iframe title="Jetzt {{ .Title }} hören." class="lazy-load" frameborder="0" height="200px" scrolling="no" seamless src="" data-src="https://hearthis.at/embed/{{ .Params.embed }}/transparent_black/?hcolor=&color=&style=2&block_size=2&block_space=1&background=1&waveform=0&cover=1&autoplay=0&css=" width="100%"></iframe>
<noscript>
    <iframe title="Jetzt {{ .Title }} hören." frameborder="0" height="200px" scrolling="no" seamless src="https://hearthis.at/embed/{{ .Params.embed }}/transparent_black/?hcolor=&color=&style=2&block_size=2&block_space=1&background=1&waveform=0&cover=1&autoplay=0&css=" width="100%"></iframe>
</noscript>
```


jpetto:
Here's the network log:

Screen Shot 2019-04-30 at 10 52 50 AM

I _think_ all the resources will be covered under `child-src`, as it's the `<iframe>` loading them. But maybe I'm wrong?

metadave:
> Will *.heartthis.at work for both cases above?

It seems like `*.heartthis.at` wouldn't catch `https://heartthis.at`

jpetto:
> It seems like `*.heartthis.at` wouldn't catch `https://heartthis.at`

Okay - I wasn't sure how generous that `*` would be. I'll update with another rule.

metadave:
@jpetto I'm fine applying the Terraform to see if it works for both domains.

metadave:
terraform applied

PR #1107, opened 2019-05-10
Allow player.simplecast.com on IRL

stephaniehobson:


metadave:
:+1:, Terraform applied

PR #1158, opened 2019-08-21
Added snippets cdn

limed:
Snippet CDN terraform code. This works with terraform 0.11.x


jgmize:
Worked with @limed to finish this up (thanks again!). This has now been applied and the next step is to schedule to move with AWS.

Issue #1042, opened 2019-02-26
transfer snippets CDN to MozMEAO AWS account


metadave:
@mozmeao/theconn let's talk about moving the snippets Cloudfront distribution from IT to the MozMEAO AWS account. MDN worked with AWS support to reuse the domain name, I think we'd need to do the same.

We would need:

- [ ] Terraform automation
- [ ] test plan
- [ ] stage CDN
- [ ] prod CDN

anything else?

https://github.com/mdn/infra/issues/112
https://github.com/mdn/infra/issues/43

cc @bensternthal

metadave:
This may impact snippets migration to GKE: https://github.com/mozmeao/infra/issues/1041

jgmize:
I think we should also discuss the cost, benefits, and estimated level of effort of migrating snippets to a new cloudfront distribution vs another CDN provider, such as Cloudflare.

metadave:
> I think we should also discuss the cost, benefits, and estimated level of effort of migrating snippets to a new cloudfront distribution vs another CDN provider, such as Cloudflare.

+1

glogiotatidis:
+1 I wanted to do this for a while!

Somewhat related is that I'm working on things like https://github.com/mozmeao/snippets-service/issues/841 which I expect to significantly reduce the CDN transfer costs.

jgmize:
This issue is also being tracked in https://jira.mozilla.com/browse/SRE-313 which is a medium priority issue in the [current IT SRE sprint](https://jira.mozilla.com/secure/RapidBoard.jspa?rapidView=95&view=planning) and @limed and I hope to find some time to work together on it this week.

glogiotatidis:
@jgmize could you please update on status?

jgmize:
@limed & I have some time on the calendar to work on this tomorrow morning. Next actions:
- [ ] set up a cert via ACM valid for `snippets.cdn.mozilla.net` and `snippets-prod-cdn.moz.works`
  - [ ] add DNS entry required for validation of domain cert for ACM
- [ ] update https://github.com/mozmeao/infra/pull/1158/files with the ARN of the new cert
- [ ] `terraform apply`
- [ ] add `snippets-prod-cdn.moz.works` DNS entry
- [ ] test distribution
- [ ] schedule move with AWS support

jgmize:
@glogiotatidis see above and https://github.com/mozmeao/infra/pull/1158#issuecomment-526343628 

[...]

PR #1159, opened 2019-09-04
snippets lambda

metadave:
This PR renames the `cdn` dir to `prod_cdn` and `stage_cdn` so that Lambda@Edge functions aren't deployed to prod, as it's now [in use](https://github.com/mozmeao/infra/issues/1042#issuecomment-528045185). Once we're ready to deploy to prod, the `stage_cdn` dir can be removed, and `prod_cdn` can be renamed to `cdn`.

@glogiotatidis I used one of the example Python lambda functions from [here](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-examples.html), but it's returning a 503 for some reason. I'll try to troubleshoot a bit more later in the week.

```
curl -I https://d177vs5g6znz6f.cloudfront.net/foobar123
HTTP/1.1 503 Service Unavailable
```


metadave:
https://github.com/mozmeao/infra/pull/1160

PR #1160, opened 2019-09-04
snippets lambda

metadave:
This PR renames the `cdn` dir to `prod_cdn` and `stage_cdn` so that Lambda@Edge functions aren't deployed to prod, as it's now [in use](https://github.com/mozmeao/infra/issues/1042#issuecomment-528045185). Once we're ready to deploy to prod, the `stage_cdn` dir can be removed, and `prod_cdn` can be renamed to `cdn`.

@glogiotatidis I used one of the example Python lambda functions from [here](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-examples.html), but it's returning a 503 for some reason. I'll try to troubleshoot a bit more later in the week.

```
curl -I https://d177vs5g6znz6f.cloudfront.net/foobar123
HTTP/1.1 503 Service Unavailable
```


apps/snippets/infra/global/lambda.py
```diff
@@ -0,0 +1,15 @@
+def lambda_handler(event, context):
+     response = event['Records'][0]['cf']['response']
```
glogiotatidis: This is a Origin Response example, that's why it wasn't working. I've made some changes via the console testing out some things. Once I get comfortable with lambda@edge i'll update terraform

metadave: that's what I get for copy/pasting code :-) thanks for looking into this @glogiotatidis!

glogiotatidis:
Can I have a couple of bullets on how to tf apply this?

metadave:
@glogiotatidis sent you a link to a gdrive dir that has a doc + video on how to get things running.

glogiotatidis:
@metadave Updated this branch with the current lambda setup and with a commit to fix #1167 . This is applied on production

glogiotatidis:
Also again thanks for the toolkit, really helpful!

PR #1178, opened 2019-10-29
Remove snippets lambda add custom 403

glogiotatidis:
This is for stage only and has been applied. 

See also https://github.com/mozmeao/snippets-service/issues/1112 and https://github.com/mozmeao/snippets-service/pull/1210


glogiotatidis:
Added configuration for 403 in production and applied.

PR #1231, opened 2020-03-04
Adding logging to the snippets elb

duallain:


tf/us-west-2/elbs.tf
```diff
@@ -264,6 +264,12 @@ resource "aws_elb" "snippets-admin" {
     "KubernetesCluster" = "oregon-b.moz.works"
     "Stack"             = "snippets-admin"
   }
+
+  access_logs {
+    bucket        = "mozmeao-elb-access-logs-us-west-2"
+    interval      = 60
```
metadave: are we ok with 60 minutes?

duallain: it looks like that's the default. I originally thought it was seconds, an hour does seem a little long.  I'll drop it to 15 minutes.

PR #1232, opened 2020-03-04
first draft of prometheus adr

duallain:


glogiotatidis:
nice write up, small, simple, precise. Helps us communicate async better. Thanks for doing this!

PR #1233, opened 2020-03-04
changing adr status

duallain:


duallain:
I believe the correct process is to mark these as Accepted, but I left them in draft.  This change is just to keep the adrs up to date with what I believe the process is.

metadave:
this is fine, I'm not too concerned about the status as we don't have a ton of ADRs at this point.

Regarding approval, we'll often have discussed an issue as a team before it becomes an ADR, so a simple +1 from anyone is fine in that case. If the ADR is a proposal and you're trying to get buy-in from the team, mention that on the PR and we'll try to reach concensus.

PR #1256, opened 2020-04-07
ADR to record a simple network design

duallain:
Attempting to document CIDR blocks to use for vpc and subnets.

### DIFF
--- a/docs/architecture/decisions/0005-vpc-and-network-design.md
+++ b/docs/architecture/decisions/0005-vpc-and-network-design.md
@@ -1,0 +1,43 @@
+# 5. VPC and Network Design
+
+Date: 2020-04-07
+
+## Status
+
+Accepted
+
+## Context
+
+Changing networking can be hard.  It usually requires a full redeploy of all services and all infrastructure to make it 'real'.  Our current network has overlaps, which makes it more difficult to setup vpns, usually requring syncing of live IP addresses to their desired state.
+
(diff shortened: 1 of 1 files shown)


docs/architecture/decisions/0005-vpc-and-network-design.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Context
+
+Changing networking can be hard.  It usually requires a full redeploy of all services and all infrastructure to make it 'real'.  Our current network has overlaps, which makes it more difficult to setup vpns, usually requring syncing of live IP addresses to their desired state.
+
+This document is referenced for design, in particular the sizing section which says, "The majority of AWS customers use VPCs with a /16 netmask and subnets with /24 netmasks". https://aws.amazon.com/answers/networking/aws-single-vpc-design/
+
+## Decision
+
+For each VPC make a block of /16 ipv4 addresses.  Where VPC maps to a region within a cloud provider.  Divide that into /24 subnets, where we'll have just one subnet per AZ.
```
metadave, line 17: the block should be unique across all of our cloud providers.

duallain:
> +1, thanks for sorting this. Will we keep a spreadsheet or doc somewhere that lists what clusters are using which blocks (so we don't accidentally reuse the same block in 2 different regions etc)?

I've added a note to the bottom of the adr that points at a couple of places I think would be ok to keep documentation. (and made sure the aws portion was up to date).

PR #1260, opened 2020-04-09
Adding an elb for nucleus dev

duallain:
Manually created acm cert for it.
Then followed prod pattern.
Launch config changes were added to keep from changing what's deployed.


duallain:
yeah, I changed what was in terraform until it showed that those resources wouldn't be changed.

PR #1332, opened 2020-07-08
Add script to get S3 bucket size.

glogiotatidis:


tools/aws/s3_bucket_size.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+#
+# Queries Cloud Watch to get the size of each S3 bucket in the account.
+#
+set -euo pipefail
+
+YESTERDAY=$(date +%Y-%m-%d --date=yesterday)
```
duallain, line 7: Had some trouble with this (for whatever reason) on my mac.  Does this work on your machine as well?

```YESTERDAY=$(date -v -1d '+%Y-%m-%d')```

glogiotatidis: No doesn't work for me. I'm using GNU coreutils `date`.

duallain: what do you think about something like this:

```if [ $( uname -s ) == "Darwin" ]
then
    YESTERDAY=$( date -v -1d '+%Y-%m-%d' )
else
    YESTERDAY=$( date +%Y-%m-%d --date=yesterday )
fi```

glogiotatidis: updated

PR #1344, opened 2020-08-27
Updating elbs to point at new backend nodes

duallain:
lbs are added to asg (which seems pretty nice!)
needed to update the k8s nodes, which were presenting some issues


duallain:
This was deployed, so merging to try to keep main branch in sync with reality.

PR #1345, opened 2020-08-28
Updated How We Work For 2020

bensternthal:
This is a first shot at updating how we work, significant changes include:
- trying to use a single board 
- not using milestones
- using just labels for epics
- organizing/committing to work in a 2 week sprint.

@glogiotatidis @pmac please take a look and let's collaborate on changes / tweaks.

### DIFF
--- a/docs/how_we_work.md
+++ b/docs/how_we_work.md
@@ -1,65 +1,55 @@
-# How we work in 2018
+# How we work in 2020
 
-#### tl;dr 
+Our team manages it's work via Github [projects](https://github.com/orgs/mozmeao/projects/2). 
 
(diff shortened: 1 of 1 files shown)


glogiotatidis:
Looks good. 

I wonder if it makes sense to use GH Milestones as Epics so we can group tasks.

pmac:
Looks good to me too. GH Milestones are nice for a single-repo project, but don't work if you want to look at tasks across multiple repos. They might be good enough though with minimal cross referenced tasks.

bensternthal:
Im up for suggestions on how to use GH milestones :)

PR #1369, opened 2021-01-28
Removing a bunch of items that shouldn't be needed after the move out…

duallain:
… of kops


glogiotatidis (approved):
Scanning the list of deleted files I agree we shouldn't be needing any of these going fwd. Thanks for cleaning things up Alan!

